PR #38, opened 2021-12-02
Support numbers in anchors

mishig25:
Currently numbers are stripped from anchor links.
This PR fixes this issue

cc: @stas00


sgugger (approved):
Thanks for adding this!

PR #39, opened 2021-12-03
Add Arguments: to the supported parameters introduction

sgugger:
This fixes the issue with some parameters not being recognized because they are introduced with "Arguments:" (cf problem [here](https://github.com/huggingface/transformers/issues/14613#issuecomment-985538424)).


mishig25 (approved):
lgtm!

PR #42, opened 2021-12-07
Add colabdropdown svelte import

mishig25:
ColabDropdown needs to be a svelte component

https://user-images.githubusercontent.com/11827707/145053749-6efbaf22-b11b-4887-8767-92521fe8174c.mov


sgugger (approved):
Thanks for adding that!

PR #43, opened 2021-12-07
Add open-in-colab syntax and fixes to conversion

sgugger:
This PR adds the syntax to resolve [[open-in-colab]] in doc files to the proper svelte component.

It also introduces a few fixes used when converting some rst files.

### DIFF
--- a/src/doc_builder/commands/convert_doc_file.py
+++ b/src/doc_builder/commands/convert_doc_file.py
@@ -68,8 +68,14 @@ def convert_command(args):
     with open(source_file, "r", encoding="utf-8") as f:
         text = f.read()
 
-    page_info = {"package_name": package_name, "page": source_file.with_suffix(".html").relative_to(doc_folder)}
+    page_info = {
+        "package_name": package_name,
+        "page": source_file.with_suffix(".html").relative_to(doc_folder),
+        "no_prefix": True,
+    }
     text = convert_rst_to_mdx(text, page_info, add_imports=False)
+    text = text.replace("&amp;lcub;", "{")
+    text = text.replace("&amp;lt;", "<")
     text = re.sub(r"^\[\[autodoc\]\](\s+)(transformers\.)", r"[[autodoc]]\1", text, flags=re.MULTILINE)
 
     with open(output_file, "w", encoding="utf-8") as f:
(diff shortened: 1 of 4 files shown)


src/doc_builder/commands/convert_doc_file.py
```diff
@@ -68,8 +68,14 @@ def convert_command(args):
...
 
-    page_info = {"package_name": package_name, "page": source_file.with_suffix(".html").relative_to(doc_folder)}
+    page_info = {
+        "package_name": package_name,
+        "page": source_file.with_suffix(".html").relative_to(doc_folder),
+        "no_prefix": True,
+    }
     text = convert_rst_to_mdx(text, page_info, add_imports=False)
+    text = text.replace("&amp;lcub;", "{")
+    text = text.replace("&amp;lt;", "<")
```
sgugger, line 78: When converting back those Markdown files, this will be undoe.

src/doc_builder/commands/convert_doc_file.py
```diff
@@ -68,8 +68,14 @@ def convert_command(args):
     with open(source_file, "r", encoding="utf-8") as f:
         text = f.read()
 
-    page_info = {"package_name": package_name, "page": source_file.with_suffix(".html").relative_to(doc_folder)}
+    page_info = {
+        "package_name": package_name,
+        "page": source_file.with_suffix(".html").relative_to(doc_folder),
+        "no_prefix": True,
```
sgugger, line 74: We don't want the whole prefix with master and language when resolving links to other doc pages or references in the source files.

src/doc_builder/commands/convert_doc_file.py
mishig25: could you elaborate on this?

sgugger: The source files in the doc should not have those special chars as we don't expect the users to write the doc this way (we're inside the command that converts an rst to the new expected doc format). We apply the conversion for those files in the `doc_builder` in [this function](https://github.com/huggingface/doc-builder/blob/66c63fdc291717fc40f0c2b19aa87390f6a5aafa/src/doc_builder/convert_md_to_mdx.py#L39).

mishig25: oh I see now

mishig25 (approved):
lgtm!

sgugger:
I think we're safe until there is an actual `[[open-in-colab]]` in the source files on the Transformers side, no?

mishig25:
that is true 👍

PR #44, opened 2021-12-08
Add custom path for build and morehelpful error message.

sgugger:
This PR adds two things:
- first it allows the user to specify where they want the doc to be built (if they want a custom dir)
- secondly, it raises more helpful error when there is something wrong: it indicates which file caused the problem and if the problem comes from an object not found, it shows which one.


src/doc_builder/autodoc.py
```diff
@@ -239,10 +239,13 @@ def document_object(object_name, package, page_info, full_name=True):
     if "package_name" not in page_info:
         page_info["package_name"] = package.__name__
     obj = find_object_in_package(object_name=object_name, package=package)
+    if obj is None:
+        raise ValueError(
+            f"Unable to find {object_name} in {package.__name__}. Make sure the path to that object is correct."
```
LysandreJik, line 244: Very nice

LysandreJik (approved):
This looks great to me. Should be much simpler to debug things from now on!

mishig25 (approved):
lgtm! thanks for adding those
The newly added error message would be super valuable

PR #45, opened 2021-12-09
Manual fix youtube

mishig25:
Manually fixing it so that the changes are up ASAP.
Will work on changing the code

cc: @sgugger @LysandreJik


sgugger:
Thanks for fixing!

PR #48, opened 2021-12-09
Convert doc page to notebooks

sgugger:
This PR adds the ability to convert doc pages to notebook auto-magically, as long as they have the `[[open-in-colab]]` marker: when building the documentation, a `notebook_dir` just has to be provided.

The generated notebooks look exactly like their doc page counterpart with a few exceptions:
- the code samples are parsed in proper code cells, and the doctest syntax is removed to separate inputs and outputs.
- the youtube svelte components are replaced by hidden cells displaying the videos
- all links are treated to point to the doc (links to an object in the package lie [`Trainer`] are of course properly resolved)
- LaTeX special syntax is changed to the usual $ $ 
- PT/TF content is split to generate one Mixed, one PyTorch-only and one TF-only notebook.

To test locally:
```
doc-builder build transformers transformers/docs/source/ --build_dir test-build --notebook_dir notebooks
```
(adapt the path to the Transformers docs source folder if necessary).

Note that notebooks generated need to be trusted before the videos are displayed properly (in colab it should be automatic but when using a jupyter notebook server, click File and Trust).

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -335,3 +335,44 @@ def autodoc(object_name, package, methods=None, return_anchors=False, page_info=
     documentation = '<div class="docstring">\n' + documentation + "</div>\n"
 
     return (documentation, anchors) if return_anchors else documentation
+
+
+def resolve_links_in_text(text, package, mapping, page_info):
+    """
+    Resolve links of the form [`SomeClass`] to the link in the documentation to `SomeClass`.
+
+    Args:
+    - **text** (`str`) -- The text in which to convert the links.
+    - **package** (`types.ModuleType`) -- The package in which to search objects for.
+    - **mapping** (`Dict[str, str]`) -- The map from anchor names of objects to their page in the documentation.
+    - **page_info** (`Dict[str, str]`) -- Some information about the page.
+    """
+    package_name = page_info.get("package_name", package.__name__)
+    version = page_info.get("version", "master")
+    language = page_info.get("language", "en")
+
+    prefix = f"/docs/{package_name}/{version}/{language}/"
+
+    def _resolve_link(search):
+        object_name = search.groups()[0]
+        # If the name begins with `~`, we shortcut to the last part.
+        if object_name.startswith("~"):
+            obj = find_object_in_package(object_name[1:], package)
+            object_name = object_name.split(".")[-1]
+        else:
+            obj = find_object_in_package(object_name, package)
+        if obj is None:
+            return f"`{object_name}`"
+
+        # If the object is not a class, we add ()
+        if not isinstance(obj, type):
+            object_name = f"{object_name}()"
+
+        # Link to the anchor
+        anchor = get_shortest_path(obj, package)
+        if anchor not in mapping:
+            return f"`{object_name}`"
+        page = f"{prefix}{mapping[anchor]}"
+        return f"[{object_name}]({page}#{anchor})"
+
+    return re.sub(r"\[`([^`]+)`\]", _resolve_link, text)
--- a/src/doc_builder/build_doc.py
+++ b/src/doc_builder/build_doc.py
@@ -22,9 +22,10 @@ from pathlib import Path
 
 from tqdm import tqdm
 
-from .autodoc import autodoc, find_object_in_package, get_shortest_path, remove_example_tags
+from .autodoc import autodoc, remove_example_tags, resolve_links_in_text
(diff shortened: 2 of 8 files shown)


src/doc_builder/autodoc.py
```diff
@@ -335,3 +335,44 @@ def autodoc(object_name, package, methods=None, return_anchors=False, page_info=
     documentation = '<div class="docstring">\n' + documentation + "</div>\n"
 
     return (documentation, anchors) if return_anchors else documentation
+
+
+def resolve_links_in_text(text, package, mapping, page_info):
```
sgugger, line 340: Moved this to the `autodoc` module to avoid a circular import in `build_doc`

src/doc_builder/build_doc.py
```diff
@@ -317,24 +277,67 @@ def generate_frontmatter(doc_folder):
...
+    doc_folder = Path(doc_folder)
+
+    if "package_name" not in page_info:
+        page_info["package_name"] = package.__name__
+
+    mdx_files = list(doc_folder.glob("**/*.mdx"))
+    for file in tqdm(mdx_files, desc="Building the notebooks"):
+        with open(file, "r", encoding="utf-8") as f:
+            if "[[open-in-colab]]" not in f.read():
+                continue
```
sgugger, line 304: We only convert doc pages that are supposed to have a notebook.

src/doc_builder/build_doc.py
```diff
@@ -317,24 +277,67 @@ def generate_frontmatter(doc_folder):
...
     Args:
-    - **package_name** (`str`) -- The name of the package.
-    - **doc_folder** (`str` or `os.PathLike) -- The folder in which the source documentation of the package is.
-    - **output_dir** (`str` or `os.PathLike) -- The folder in which to put the built documentation. Will
-      be created if it does not exist.
-    - **clean** (`bool`, *optional*, defaults to `True`) -- Whether or not to delete the content of the
-      `output_dir` if that directory exists.
-    - **version** (`str`, *optional*, defaults to `"master"`) -- The name of the version of the doc.
-    - **language** (`str`, *optional*, defaults to `"en"`) -- The language of the doc.
+        package_name (`str`): The name of the package.
```
sgugger, line 320: Slowly migrating to the format we said we would keep for the docs :-)

src/doc_builder/convert_to_notebook.py
```diff
@@ -0,0 +1,301 @@
...
+    language = page_info.get("language", "en")
+    page = str(page_info["page"])
+
+    prefix = f"https://huggingface.co/docs/{package_name}/{version}/{language}"
+
+    def _replace_link(match):
+        description, link = match.groups()
+        if link.startswith("http") or link.startswith("#"):
+            return f"[{description}]({link})"
+        elif link.startswith("/docs/"):
```
sgugger, line 54: Links resolved for doc objects are of this form.

tests/test_build_doc.py
```diff
@@ -47,104 +40,6 @@ def test_resolve_open_in_colab(self):
 """
         self.assertEqual(resolve_open_in_colab("\n[[open-in-colab]]\n", {"page": "quicktour.html"}), expected)
 
-    def test_resolve_links_in_text(self):
```
sgugger, line 50: Just moved those tests in the proper test file with the move of that function.

src/doc_builder/autodoc.py
```diff
@@ -335,3 +335,44 @@ def autodoc(object_name, package, methods=None, return_anchors=False, page_info=
...
+    - **text** (`str`) -- The text in which to convert the links.
+    - **package** (`types.ModuleType`) -- The package in which to search objects for.
+    - **mapping** (`Dict[str, str]`) -- The map from anchor names of objects to their page in the documentation.
+    - **page_info** (`Dict[str, str]`) -- Some information about the page.
+    """
+    package_name = page_info.get("package_name", package.__name__)
+    version = page_info.get("version", "master")
+    language = page_info.get("language", "en")
+
+    prefix = f"/docs/{package_name}/{version}/{language}/"
```
LysandreJik, line 354: Probably not important, but this won't work on Windows (I know it was here before!)

LysandreJik (approved):
Nice! Really cool to see some notebook magic here :star_struck:

src/doc_builder/autodoc.py
sgugger: It's an URL not a local path. I think they work the same way on Windows ;-)

PR #50, opened 2021-12-10
Add custom first cell

sgugger:
This PR adds the ability to customize the first cell(s) of auto-generated notebooks. For transformers we want to install `transformers` and `datasets`  but for other packages we will want something different.

I took the approach of using a `_config.py` file (a bit like the old conf.py for Sphinx or the one used in pytest) that would look like this for Transformers (see [this PR](https://github.com/huggingface/transformers/pull/14718)):
```
INSTALL_CONTENT = """
# Transformers installation
! pip install transformers datasets
# To install from source instead of the last release, comment the command above and uncomment the following one.
# ! pip install git+https://github.com/huggingface/transformers.git
"""

notebook_first_cells = [{"type": "code", "content": INSTALL_CONTENT}]
```

And since we might add other customizable things in the future, I made the architecture to get that config as a Python module.

### DIFF
--- a/src/doc_builder/build_doc.py
+++ b/src/doc_builder/build_doc.py
@@ -27,6 +27,7 @@ from .convert_md_to_mdx import convert_md_to_mdx
 from .convert_rst_to_mdx import convert_rst_to_mdx, find_indent, is_empty_line
 from .convert_to_notebook import generate_notebooks_from_file
 from .frontmatter_node import FrontmatterNode
+from .utils import read_doc_config
 
 
 _re_autodoc = re.compile("^\s*\[\[autodoc\]\]\s+(\S+)\s*$")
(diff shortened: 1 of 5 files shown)


LysandreJik (approved):
Very nice to have it as a configurable option. This should make adoption in other libraries simpler!

MKhalusova:
@sgugger Is the first notebook cell customizable on the library level, or can it be customized for individual doc pages?
I noticed that at least a couple of notebooks were requiring `evaluate` in addition to `transformers` and `datasets`. Here are a couple of examples: 
https://huggingface.co/docs/transformers/tasks/sequence_classification
https://huggingface.co/docs/transformers/tasks/semantic_segmentation

sgugger:
The policy is just to add it in all notebooks (like `datasets` which is not used in all doc pages for instance), so I'd just add `evaluate` in that config. `evaluate` should also be used everywhere in the doc instead of `datasets` for metric computation at this stage :-)

MKhalusova:
@sgugger Ok, I will create a PR to transformers to include `evaluate` in the generated cell.

MKhalusova:
I've opened a PR for this https://github.com/huggingface/transformers/pull/20850

PR #54, opened 2021-12-13
Add a check for the ToC

sgugger:
This PR adds a check for the ToC, specifically that it contains all the doc-generated files. If it does not, the `doc-builder` will now error (like sphinx used to).

As a result this PR shouldn't be merged without its [counterpart](https://github.com/huggingface/transformers/pull/14755) on the Transformers side (which adds missing files in the ToC).


LysandreJik (approved):
Very clear! LGTM!

PR #55, opened 2021-12-13
Don't error if package has no doc config

sgugger:
As seen in Transformers earlier today, the doc building will fail if there is no config file in the doc folder, which shouldn't be the case since users are allowed to not want to customize their doc building.


LysandreJik (approved):
Good catch!

PR #61, opened 2021-12-17
Add docstrings conversion to doc-builder convert

sgugger:
This PR adds a command to convert all the docstrings in a python file to Markdown. It uses the already existing command `doc-builder convert`, which has a different behavior for rst files (convert them to MDX) and py files (convert all the docstrings inside that are written in rst).

You can test this in the Transformers repo by running:
```bash
doc-builder convert src/transformers/models/bert/modeling_bert.py
make style
```
(make style is to re-organize the content with the 119 char limit) and enjoy the diff!

### DIFF
--- a/src/doc_builder/commands/convert_doc_file.py
+++ b/src/doc_builder/commands/convert_doc_file.py
@@ -18,7 +18,39 @@ import argparse
 import re
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


LysandreJik (approved):
Very nice! I should use it extensively sometime today.

PR #64, opened 2021-12-20
Docstring MD to MDX conversion fixes

sgugger:
This PR fixes a little bit the conversion MD to MDX by removing the indent (that stays for all the arguments otherwise). It's all whitespace, so not sure it has a real impact on moon-landing, but it removes the diff I get when converting docstrings of the models from rst to MarkDown.


mishig25 (approved):
Lgtm!

PR #68, opened 2021-12-22
Add defensive checks to prevent moon-landing deployment issues

sgugger:
This PR adds a few defensive checks to make sure the docstrings built do not cause any issue on moon-landing.

It needs some fixes on the Transformers side so do not merge yet :-)


sgugger:
Will merge as soon as this is green so that we don't get the same mess as yesterday night again :-) 
Happy to address any comment in a followup PR!

LysandreJik:
Perfect! Thanks for taking care of this!

PR #75, opened 2022-01-11
Dev build should not update _versions.yml file

mishig25:
Dev build should not update _versions.yml file

Reusing this bool/flag https://github.com/huggingface/doc-builder/blob/0e9fffb822ef46edc5aa10c0f44baa6193e41732/src/doc_builder/commands/build.py#L29

We can add other agument like `if arg PR_number exists` wdyt

### DIFF
--- a/src/doc_builder/commands/build.py
+++ b/src/doc_builder/commands/build.py
@@ -46,7 +46,7 @@ def build_command(args):
         notebook_dir=args.notebook_dir,
     )
 
-    # dev build should not uodate _versions.yml
+    # dev build should not update _versions.yml
     package_doc_path = os.path.join(args.build_dir, args.library_name)
     if "dev" not in version and os.path.isfile(os.path.join(package_doc_path, "_versions.yml")):
         update_versions_file(os.path.join(args.build_dir, args.library_name), version)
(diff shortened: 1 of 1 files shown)


src/doc_builder/commands/build.py
```diff
@@ -46,8 +46,9 @@ def build_command(args):
         notebook_dir=args.notebook_dir,
     )
 
+    # dev build should not uodate _versions.yml
     package_doc_path = os.path.join(args.build_dir, args.library_name)
-    if os.path.isfile(os.path.join(package_doc_path, "_versions.yml")):
+    if "dev" not in version and os.path.isfile(os.path.join(package_doc_path, "_versions.yml")):
```
sgugger, line 51: Let's see what we use for the "version" when pushing PRs builds from transformers. This assumes there is dev in it, but I was thinking of `pr_xxx` with xxx being the PR number.

mishig25: sounds good to me

mishig25:
I think this is the only change we need for now to make the dev setup working

src/doc_builder/commands/build.py
sgugger: Ok so it looks like we are going with version being pr_xxx so can you change the dev here to pr?

PR #77, opened 2022-01-26
Fix real links with a code block

sgugger:
Currently links with a code block like \[\`Accelerator\`\]\(https://huggingface.co/docs/accelerate/accelerator.html#accelerate.Accelerator\) are changed by `doc-builder` when it attemps to resolve the links to other parts of the docs (since it does not found the object in the doc of Transformers). This PR fixes that and adds a test.

### DIFF
--- a/tests/test_autodoc.py
+++ b/tests/test_autodoc.py
@@ -354,6 +354,16 @@ tuple before.
             ),
         )
 
+        self.assertEqual(
+            resolve_links_in_text(
+                "This is a regular [`link`](url)",
+                transformers,
+                small_mapping,
+                page_info,
+            ),
+            "This is a regular [`link`](url)",
+        )
+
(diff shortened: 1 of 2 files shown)


tests/test_autodoc.py
```diff
@@ -354,6 +354,16 @@ def test_resolve_links_in_text(self):
...
         )
 
+        self.assertEqual(
+            resolve_links_in_text(
+                "This is a regular [`link`](url)",
+                transformers,
+                small_mapping,
+                page_info,
+            ),
+            "This is a regular [`link`](url)",
```
sgugger, line 364: This test was failing, the returned value used to be "This is a regular \`link\`(url)"

mishig25 (approved):
lgtm!

PR #79, opened 2022-01-31
Only add version if it's not there already

sgugger:
This makes sure the version is not added to `_versions.yml` if it's already there, to avoid duplicates (like in [this commit](https://github.com/huggingface/doc-build/commit/f3c1b81c9967d09cf0f1c56c80977aca2bb8ece0)).


mishig25 (approved):
lgtm!

PR #80, opened 2022-01-31
Add test case for `fix_duplicate_versions`

mishig25:


sgugger:
Thanks for adding a test!

PR #85, opened 2022-02-09
Ignore property in docstrings

mishig25:
Currently, ignoring if a method is @property method. Will need to handle it in a separate repo since we cannot use `inspect` on `property` type 
[internal discussion](https://huggingface.slack.com/archives/C02GLJ5S0E9/p1644339518387909)


sgugger:
I think #87 is a better fix.

mishig25:
@sgugger perffect!

PR #86, opened 2022-02-09
Catch raises section in docstring

mishig25:
Unlike transformers docs, datasets docs has a lot of using `raises` section in docstring. Therefore, this PR implements catching raises sections in docstrings

### DIFF
--- a/src/doc_builder/convert_rst_to_mdx.py
+++ b/src/doc_builder/convert_rst_to_mdx.py
@@ -440,9 +440,10 @@ def parse_rst_docstring(docstring):
             # The line may contain the return type.
             return_type, lines[idx] = split_return_line(lines[idx])
             idx += 1
-            while idx < len(lines) and_re_raises.search(lines[idx]) is None and (
-                is_empty_line(lines[idx])
-                or find_indent(lines[idx]) >= return_indent
+            while (
+                idx < len(lines)
+                and _re_raises.search(lines[idx]) is None
+                and (is_empty_line(lines[idx]) or find_indent(lines[idx]) >= return_indent)
             ):
                 idx += 1
             lines.insert(idx, "</returns>\n")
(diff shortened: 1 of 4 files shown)


src/doc_builder/convert_rst_to_mdx.py
```diff
@@ -394,6 +396,34 @@ def parse_rst_docstring(docstring):
             lines.insert(idx, "</parameters>\n")
             idx += 1
 
+        # Raises section
```
sgugger, line 399: I can't comment on line 385 but the test there will need to include the `_re_raises` as well now, same for the test at line 440.

Might be worth grouping both the "Returns" and "Raises" in the same re pattern and then have subtest there? I'll let you decide.

sgugger (approved):
Thanks for working on this!

src/doc_builder/convert_rst_to_mdx.py
```diff
@@ -437,7 +440,11 @@ def parse_rst_docstring(docstring):
             # The line may contain the return type.
             return_type, lines[idx] = split_return_line(lines[idx])
             idx += 1
-            while idx < len(lines) and (is_empty_line(lines[idx]) or find_indent(lines[idx]) >= return_indent):
+            while idx < len(lines) and (
+                is_empty_line(lines[idx])
+                or find_indent(lines[idx]) >= return_indent
+                or _re_raises.search(lines[idx]) is not None
+            ):
```
mishig25, line 447: @sgugger let me know if the brackets look good 😊

sgugger: Ah, thanks for checking! It is not ;-)
Here is what we need:
```suggestion
            while idx < len(lines) and_re_raises.search(lines[idx]) is None and (
                is_empty_line(lines[idx])
                or find_indent(lines[idx]) >= return_indent
            ):
```

mishig25: thanks a lot!

PR #87, opened 2022-02-09
Make `autodoc` with properties

sgugger:
This PR makes `autodoc` work with properties. The trick is to replace them with their getter, which is a normal Python function.


mishig25 (approved):
lgtm!

PR #90, opened 2022-02-13
Escape } with &amp;rcub;

mishig25:


mishig25:
Actually, this change is not needed

PR #92, opened 2022-02-14
Add `Attributes` to `_re_args`

mishig25:
Catch `Attributes` keyword in `_re_args` because of [this](https://github.com/huggingface/datasets/blob/21bfd0d3f5ff3fbfd691600e2c7071a167816cdf/src/datasets/info.py#L95-L101)


sgugger (approved):
Thanks for adding this!

PR #94, opened 2022-02-14
Generate HTML docs

coyotte508:
What's left:

- [ ] CSS : No CSS/Tailwind is generated. It was done in [`0a2b7c8` (#94)](https://github.com/huggingface/doc-builder/pull/94/commits/0a2b7c804392a52dfa73be674ff1c041e5038b7e) but reverted, because moonlanding should have the css already. But does it have all of it?
- [ ] Test : make sure everything works fine
- [x] Update the CI in transformers: https://github.com/huggingface/transformers/pull/15678
- [ ] Update https://moon-ci-docs.huggingface.co/docs/transformers/
- [x] Instead of generating `en.html`, try generating `en/index.html` (done in the post-build script. Note this might be a svelte bug, [pages should generate an index.html](https://kit.svelte.dev/docs/page-options#prerender-route-conflicts)) 
- [ ] Update old docs in https://github.com/huggingface/doc-build  : https://github.com/huggingface/doc-build/pull/2
    - [ ] Docs for master
    - [x] Docs for v4.16.2 
    - [x] Docs for v4.15.0
    - [x] Docs for v4.14.1
    - [x] Docs for v4.13.0
- [ ] Update dev docs in https://github.com/huggingface/doc-build-dev - can be done by updating the transformer PRs with transformer's master after the merge
- [x] Simplify process? Currently the `.mdx` need to be generated by the python code, then the output needs to be pasted under `kit/src/routes`, and `npm run build` needs to be called . Add `--html` flag to python command? https://github.com/huggingface/doc-builder/pull/94#issuecomment-1042103792
- [ ] ...


kit/mdsvex.config.js
```diff
@@ -0,0 +1,14 @@
...
+
+	smartypants: {
+		dashes: 'oldschool'
+	},
+
+	remarkPlugins: [],
+	rehypePlugins: []
+});
+
+export default config;
```
Pierrci, line 14: is this file actually used currently?

Pierrci (commented):
I feel like it's almost too easy! (the hard part might actually be to find the correct configuration?) 😃

Did you have to modify the copied components in `/lib`?

Other than that, I agree w/ all your points!

> Simplify process? Currently the .mdx need to be generated by the python code, then the output needs to be pasted under kit/src/routes, and npm run build needs to be called

Is the plan to handle the copy/paste in a step of the github action in the Transformers CI at the beginning?

kit/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+	"useTabs": true,
+	"singleQuote": true,
```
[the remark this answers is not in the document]
coyotte508: I just used the default generated file :)

kit/mdsvex.config.js
coyotte508: Nope :octopus:

kit/.prettierrc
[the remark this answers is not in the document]
coyotte508: Done in [`09a97ef` (#94)](https://github.com/huggingface/doc-builder/pull/94/commits/09a97ef0aadaca468279ff25d6699b8610dc5298) (and actually ran prettier)

coyotte508:
> Did you have to modify the copied components in `/lib`?

Only update the imports (since I put all the files inside a single folder). And prettier.

> Is the plan to handle the copy/paste in a step of the github action in the Transformers CI at the beginning?

Yes, it won't be a problem to do it in the CI. It still feels clunky, and the "build the documentation" part of [this README](https://github.com/huggingface/transformers/tree/master/docs#readme) will be ugly.

sgugger:
The build command can copy the files and run the `npm` command (with a proper error raised at the beginning if `npm` is not installed) then copy them back to the build directory selected. Maybe with a new `--html` flag to that command, to make it optional.

Pierrci:
I tried to play the game of 7 differences, the only one I found (but there might be others haha) is some problem w/ escaping (probably), see below (cc @mishig25, it seems to be in the docstring component)

Screen Shot 2022-02-16 at 5 29 20 PM

kit/svelte.config.js
```diff
@@ -19,6 +19,7 @@ const config = {
 		},
 
 		prerender: {
+			concurrency: process.env.KIT_CONCURRENCY ? Number(process.env.KIT_CONCURRENCY) : undefined,
```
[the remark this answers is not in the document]
coyotte508: No, it was only to test the OOM bug

coyotte508:
> I tried to play the game of 7 differences, the only one I found (but there might be others haha) is some problem w/ escaping (probably), see below (cc @mishig25, it seems to be in the docstring component)
> 
> Screen Shot 2022-02-16 at 5 29 20 PM

Ok. I changed the docstring's preprocess (to remove hydrate-props, also added some `JSON.stringify`), it might be related to that

coyotte508:
The colab dropdown seems to be broken, looking into it

src/doc_builder/commands/build.py
```diff
@@ -51,6 +107,46 @@ def build_command(args):
...
+            # Copy everything in a tmp dir
+            shutil.copytree(kit_folder, tmp_dir / "kit")
+            if (tmp_dir / "kit" / "src" / "routes" / args.language).is_dir():
+                # Remove the current MDX dir if it exists.
+                shutil.rmtree(tmp_dir / "kit" / "src" / "routes" / args.language)
+            shutil.copytree(output_path, tmp_dir / "kit" / "src" / "routes" / args.language)
+
+            # Build doc with node
+            working_dir = str(tmp_dir / "kit")
+            print("Installing node dependencies")
```
sgugger, line 123: Feel free to change this statement to anything else if `npm ci` does not do that :-)

sgugger:
Pushed an `--html` flag as discussed, which performs the steps I saw in [transformers#15678](https://github.com/huggingface/transformers/pull/15678). The only thing which would need double-checking is that the `DOCS_VERSION` env variable passed is properly received on the node side, but I'm not sure what to look for here.

There are two improvements possible in follow-up PRs:
- if we see it's too much of a pain for users to run the command because the kit folder can't be located by Python, we could git clone it and cache it
- there is no progress bar for the HTML build by node. We could show the whole output of that command if that helps (not sure it does since it's so verbose).

src/doc_builder/commands/build.py
```diff
@@ -51,6 +107,47 @@ def build_command(args):
...
 
+    # If asked, convert the MDX files into HTML files.
+    if args.html:
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            tmp_dir = Path(tmp_dir)
+            # Copy everything in a tmp dir
+            shutil.copytree(kit_folder, tmp_dir / "kit")
+            if (tmp_dir / "kit" / "src" / "routes" ).is_dir():
+                # Remove the current MDX dir if it exists.
+                shutil.rmtree(tmp_dir / "kit" / "src" / "routes")
```
coyotte508, line 118: Is it possible to only remove the .mdx files?

For now there is only a `.gitkeep` file in `kit/src/routes`, but with the viewer PR there will be a `__layout.svelte` and an `endpoints` folder (containing a `.ts` file) that should stay

coyotte508:
> Pushed an `--html` flag as discussed, which performs the steps I saw in [transformers#15678](https://github.com/huggingface/transformers/pull/15678). The only thing which would need double-checking is that the `DOCS_VERSION` env variable passed is properly received on the node side, but I'm not sure what to look for here.

Thanks!!

It's passed correctly: you can see in https://github.com/huggingface/doc-build-dev/blob/main/transformers/pr_15678/en/accelerate.html links like `href="/docs/transformers/pr_15678/en/_app/start-50888a5c.js"`, the PR number is passed correctly

sgugger (approved):
Thanks again for working on this!

mishig25:
I'd be happy to bring back 0a2b7c804392a52dfa73be674ff1c041e5038b7e for the tailwind.

Tailwind works by:
1. Collect all tailwind classes by scanning all files
2. Created one bundled css file

Therefore, moon-landing is NOT guaranteed to have all the used css files. 
So there are 2 options:
1. 0a2b7c804392a52dfa73be674ff1c041e5038b7e which I favour because it is a simpler approach. Th drawback is that there might be some duplications with moon-landing for common classes like `flex`. However, I think it is fine.
2. The other approach is to have some kind whitelist like I did [here](https://github.com/huggingface/moon-landing/blob/cefa6d88f474557426a58ae9a595a463157e9bea/server/views/pages/DocPage/DocBuilderPage.svelte#L77-L101). I think this approach is hard to maintain/remember

wdyt cc: @Pierrci @gary149

Pierrci:
@mishig25 I think it depends on when/where new classes "pop up" in the doc: is it in user-generated content (i.e. are there CSS classes directly in transformers codebase for example?), or is it through our own components/processing? If it's the latter, 2. is fine - otherwise, I guess that 1 would make sense if UGC containing css classes happens often.

mishig25:
it can happen rarely directly in the codebase of libraries, for example https://github.com/huggingface/datasets/blame/cf31cf431d76c001d3e4ca6fb2202c4ca7fb771b/docs/source/index.mdx#L13-L32
Screenshot 2022-02-22 at 23 34 15

Pierrci:
Ok thanks, and when it happens it's from us it seems :).

Let's ship without tailwind then.

Pierrci (commented):
Let's merge everything and see how it goes?

Pierrci (approved):

PR #95, opened 2022-02-15
Autodocument @property

mishig25:
Most methods with @property decorator don't have any __doc__ (see [here](https://github.com/huggingface/datasets/blob/b45dd77d01486e08731a0383aaf2c6140f85576c/src/datasets/table.py#L245-L263) for an example).
And because of [this line](https://github.com/huggingface/doc-builder/blob/da20949cf0168447ba08a9940e337644fa4afc18/src/doc_builder/autodoc.py#L321), an empty doc string and doc component was being returned.

As a solution, this PR auto generates docstring for  methods with @property.

Example:

Before:
```python
@property
def num_columns(self):
    return self.table.num_columns
```

After:
```python
@property
def num_columns(self):
    """
    `@property` that returns `self.table.num_columns`
    """
    return self.table.num_columns
```

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -270,6 +270,17 @@ def get_source_link(obj, page_info):
     return f"{base_link}{module}.py#L{line_number}"
 
 
+def document_property(obj):
+    """
+    Documents property.
+    """
+    src_str = inspect.getsource(obj).strip()
+    match = re.search(r"return (.*)$", src_str)
+    returns = match.group(1)
+    obj.__doc__ = f"`@property` that returns `{returns}`"
+    return obj
+
+
 def document_object(object_name, package, page_info, full_name=True):
     """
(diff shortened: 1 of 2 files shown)


sgugger:
I'm not sure I understand the problem this PR is trying to solve. We don't want an autodoc of properties, it should be up to the user to write their docs. Looking at the documentation of the [Table class](https://huggingface.co/docs/datasets/package_reference/table_classes.html) I see none of those properties are documented by sphinx.

mishig25:
# Problem description:
When `obj.__doc__` is empty, no mdx section for that doc gets generated.

# Solution
https://github.com/huggingface/doc-builder/pull/96 fixes the problem by generating the signature str/svelte for that doc even if `obj.__doc__` is empty.

I realize that #95 was overcomplicating things. Please let me know if the problem is clear 🙂

PR #96, opened 2022-02-15
Fix document_object when objdoc is empty

mishig25:
# Problem description:
When `obj.__doc__` is empty, no mdx section for that doc gets generated.

# Solution
https://github.com/huggingface/doc-builder/pull/96 fixes the problem by generating the signature str/svelte for that doc even if `obj.__doc__` is empty.

I realize that #95 was overcomplicating things. Please let me know if the problem is clear 🙂

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -301,7 +301,6 @@ def document_object(object_name, package, page_info, full_name=True):
     name = name.replace("_", "\_")
 
     prefix = "class " if isinstance(obj, type) else ""
-    documentation = ""
     object_doc = ""
     signature_name = prefix + name
     signature = format_signature(obj)
@@ -320,7 +319,7 @@ def document_object(object_name, package, page_info, full_name=True):
 
     source_link = get_source_link(obj, page_info)
     component = get_signature_component(signature_name, anchor_name, signature, object_doc, source_link)
-    documentation += "\n" + component + "\n"
+    documentation = "\n" + component + "\n"
     return documentation, check
 
 
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -302,6 +302,7 @@ def document_object(object_name, package, page_info, full_name=True):
 
     prefix = "class " if isinstance(obj, type) else ""
     documentation = ""
+    object_doc = ""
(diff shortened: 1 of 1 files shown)


src/doc_builder/autodoc.py
```diff
@@ -317,9 +317,9 @@ def document_object(object_name, package, page_info, full_name=True):
             check = quality_check_docstring(object_doc, object_name=object_name)
             object_doc = convert_md_docstring_to_mdx(obj.__doc__, page_info)
 
-        source_link = get_source_link(obj, page_info)
-        component = get_signature_component(signature_name, anchor_name, signature, object_doc, source_link)
-        documentation += "\n" + component + "\n"
+    source_link = get_source_link(obj, page_info)
+    component = get_signature_component(signature_name, anchor_name, signature, object_doc, source_link)
+    documentation = "\n" + component + "\n"
```
mishig25, line 322: because of the if condition above and indentation, nothing was generated if `obj.__doc__` was empty.

With this change, we would see, something like this
Screenshot 2022-02-15 at 14 07 35

sgugger:
Sounds better. Just so I have the full picture, what part of the Datasets doc are you trying to fix with this?

mishig25:
All property methods have empty docstring https://github.com/huggingface/datasets/blob/b45dd77d01486e08731a0383aaf2c6140f85576c/src/datasets/table.py#L245-L263

However, they were listed to show in the doc as in https://huggingface.slack.com/archives/C02GLJ5S0E9/p1644339779420269?thread_ts=1644339518.387909&cid=C02GLJ5S0E9

sgugger:
They are not shown by sphinx (check [here](https://huggingface.co/docs/datasets/master/package_reference/table_classes.html)), so this is a change from sphinx. I'm fine with the change as you did, but I'm also fine ignoring the non-documented methods/properties like sphinx does.

Let's see what @lhoestq thinks!

lhoestq:
They should be documented as the rest. Let me add docstrings for these ones in `datasets`

I'd rather be able to see what documentation is missing in the doc, so it's maybe better if we keep them even if they don't have docstrings

sgugger:
Let's go ahead and merge this then @mishig25 :-)

PR #98, opened 2022-02-17
Preview doc in dev mode

### LINKED ISSUE #97: Add "viewer" for html docs
HTML docs are generated, and we can already view a non-interactive version by opening the HTML file.

It's possible to:

- add the css by copying the css on huggingface.co
- add interactivity by serving the html files from a webserver (from the correct folder)

Maybe we could also add a navbar with the list of html files (based on `_toctree.yml`) and have a command like `docbuilder --serve` ? It would open the docs on localhost.

It could be done through sveltekit: by running `npm run dev` , and going to http://localhost:3000/docs/transformers/master/en

coyotte508:
Fix #97 

Inside the `kit` folder, do `npm run dev`. 

Then go to http://localhost:3000/docs/transformers/master/en :  

image

This allows users to easily switch between doc files with a basic sidebar

### DIFF
--- a/kit/src/routes/__layout.svelte
+++ b/kit/src/routes/__layout.svelte
@@ -2,7 +2,6 @@
   import { base } from "$app/paths";
 
   export async function load(input: LoadInput) {
-    console.log(input.url.toString())
     const language = /\/([a-z]{2})(\/|$)/.exec(input.url.toString())[1];
 
     const toc = await input.fetch(base + '/endpoints/toc?lang='+language)
--- a/kit/src/routes/__layout.svelte
+++ b/kit/src/routes/__layout.svelte
@@ -1,0 +1,65 @@
+<script context="module" lang="ts">
+  import { base } from "$app/paths";
+
+  export async function load(input: LoadInput) {
+    console.log(input.url.toString())
+    const language = /\/([a-z]{2})(\/|$)/.exec(input.url.toString())[1];
+
+    const toc = await input.fetch(base + '/endpoints/toc?lang='+language)
+
+    return {
+      props: {
+        toc: await toc.json(),
+        language
+      }
+    }
+  }
+</script>
+
+<script lang="ts">
+  import type { LoadInput } from "@sveltejs/kit";
+  import { prerendering } from "$app/env";
+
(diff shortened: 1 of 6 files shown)


kit/src/routes/__layout.svelte
```diff
@@ -0,0 +1,64 @@
...
+
+<script lang="ts">
+  import type { LoadInput } from "@sveltejs/kit";
+  import { prerendering } from "$app/env";
+
+  export let toc: Array<{sections: Array<{title: string} & ({local: string} | {sections: {title: string, local: string}[]})>, title: string}>
+  export let language: string
+</script>
+
+{#if prerendering}
```
coyotte508, line 26: Do not render sidebar during `build`, only during `dev`

mishig25:
if we add tailwind as suggested in https://github.com/huggingface/doc-builder/pull/94#issuecomment-1047622922, the docs with css will look as good as the production version (except the left navbar)

and we can even use netlify deploy for the PR docs previews as in huggingface_hub [here](https://github.com/huggingface/huggingface_hub/runs/5085392724?check_suite_focus=true) 

cc: @Pierrci @julien-c

mishig25 (approved):
lgtm! @coyotte508 please feel free to merge it into `kit`

coyotte508:
> if we add tailwind as suggested in [#94 (comment)](https://github.com/huggingface/doc-builder/pull/94#issuecomment-1047622922), the docs with css will look as good as the production version (except the left navbar)

With [#94 (comment)](https://github.com/huggingface/doc-builder/pull/94#issuecomment-1047622922) the doc didn't look as good as on moon-landing, something about line height / spacing I think, but if we add back css I will take another look at it!

> and we can even use netlify deploy for the PR docs previews as in huggingface_hub [here](https://github.com/huggingface/huggingface_hub/runs/5085392724?check_suite_focus=true)
> 

Removing the need for moon-landing at all in the CI process would be great :) 

Maybe as a second PR though

julien-c:
> Removing the need for moon-landing at all in the CI process would be great :)

We're adding an (infra) dependency to Netlify or another moving part, I don't really see the point personally tbh

PR #101, opened 2022-02-18
Rm `name.replace("_", "\_")`

mishig25:
Fixes https://github.com/huggingface/doc-builder/pull/94#issuecomment-1042383743

@sgugger please let me know if there was a specific reason why this espacing was here initially

cc: @Pierrci


sgugger (commented):
Maybe there were problems initially since we were using the underscore for italics. But we use stars now, so should be good.

sgugger (approved):

PR #102, opened 2022-02-22
Add css to generated docs

coyotte508:
In case we want to add CSS to generated docs (https://github.com/huggingface/doc-builder/pull/94#issuecomment-1047622922):

image


kit/src/app.css
```diff
@@ -0,0 +1,422 @@
...
+}
+
+.prose.prose-doc .docstring li.text-base,
+.prose.prose-doc .docstring .text-base li {
+	@apply my-3;
+}
+
+.prose.prose-doc .docstring .text-base code {
+	@apply text-sm;
+}
```
coyotte508, line 384: Had to comment those out due to recursive tailwind errors on `text-base`:

> [vite] Internal server error: You cannot `@apply` the `text-base` utility here because it creates a circular dependency.

mishig25: @gary149 do you know of any workaround?

mishig25: in those cases, could you just write the raw css like based on the [reference](https://tailwindcss.com/docs/font-size)

for example,
Screenshot 2022-02-22 at 20 58 20

kit/src/app.css
```diff
@@ -0,0 +1,418 @@
...
+*/
```
mishig25, line 381:
```suggestion
.prose.prose-doc .docstring .text-base p {
	text-base	font-size: 1rem;
	line-height: 1.5rem;
}

.prose.prose-doc .docstring li.text-base,
.prose.prose-doc .docstring .text-base li {
	margin-top: 0.75rem;
	margin-bottom: 0.75rem;
}

.prose.prose-doc .docstring .text-base code {
	font-size: 0.875rem;
	line-height: 1.25rem;
}

```

mishig25 (approved):
lgtm!

kit/src/app.css
coyotte508: It's the class name in the selector itself I think that's causing the issue.

Even this doesn't work:

```css
.prose.prose-doc .docstring .text-base p {
	/* @apply text-base; */
	font-size: 1rem;
	line-height: 1.5rem;
}

.prose.prose-doc .docstring li.text-base,
.prose.prose-doc .docstring .text-base li {
	/* @apply my-3; */
	margin-top: 0.75rem;
	margin-bottom: 0.75rem;
}

.prose.prose-doc .docstring .text-base code {
	/* @apply text-sm; */
	font-size: 0.875rem;
	line-height: 1.25rem;
}
```

kit/src/app.css
coyotte508: Looking at the docstring component I ended up fixing it with: [`c68dda1` (#102)](https://github.com/huggingface/doc-builder/pull/102/commits/c68dda1c4fbcdcdc92f00bb4bad9c43d5ca7a659)

Only the second condition seems to be used at all (the first is not needed, since tailwind will already convert `text-base`)

coyotte508:
> lgtm!

Thanks :)

WDYT @Pierrci about adding the css in the generated html directly?

mishig25:
from @Pierrci https://github.com/huggingface/doc-builder/pull/94#issuecomment-1048304037
> Ok thanks, and when it happens it's from us it seems :).
   Let's ship without tailwind then.

lets close this PR for now, and if we need in the future, we can reopen it 😊

coyotte508:
> from @Pierrci [#94 (comment)](https://github.com/huggingface/doc-builder/pull/94#issuecomment-1048304037)
> 
> > Ok thanks, and when it happens it's from us it seems :).
> > Let's ship without tailwind then.
> 
> lets close this PR for now, and if we need in the future, we can reopen it blush

Ok. We could have kept the CSS just for the local preview but yea it would have added quite a few deps just for that.

mishig25:
@coyotte508 please feel free to repoen and change the base merging into to `master` or create a new PR' up to you 😃

coyotte508:
> @coyotte508 please feel free to repoen and change the base merging into to `master` or create a new PR' up to you smiley

I think I'm stuck - can't reopen because target is deleted, can't change target because PR is closed.

I'll create a new PR

PR #106, opened 2022-02-24
Catch docstring yields

mishig25:
Catch `yields` section in docstrings

### DIFF
--- a/src/doc_builder/convert_rst_to_mdx.py
+++ b/src/doc_builder/convert_rst_to_mdx.py
@@ -431,7 +431,7 @@ def parse_rst_docstring(docstring):
             if tag in ["return", "yield"] and return_type is not None:
                 lines[idx - 1] += f"\n<{tag}type>{return_type}</{tag}type>\n"
             elif raised_errors:
-                # rasied error
+                # raised errors
                 lines[idx - 1] += f"\n<raisederrors>{' or '.join(raised_errors)}</raisederrors>\n"
 
         else:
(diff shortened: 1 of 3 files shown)


src/doc_builder/convert_rst_to_mdx.py
```diff
@@ -431,8 +431,10 @@ def parse_rst_docstring(docstring):
 
         # Returns section
         elif _re_returns.search(lines[idx]) is not None:
+            # tag is either `return` or `yield`
```
sgugger, line 434: Could we group the `Raises` block with this? We would just need a map from `tag` to `tag_type` that sends `raises` to `rasiederrors` and `returns`/`yields` to `returnstype`/`yieldstype`.

sgugger (commented):
Thanks for working on this. Could you also add a test?

mishig25:
Thanks a lot for the suggestions! I've applied those suggestions; so please feel free to re-review them 😊

src/doc_builder/convert_rst_to_mdx.py
```diff
@@ -441,21 +408,31 @@ def parse_rst_docstring(docstring):
...
+
             lines.insert(idx, f"</{tag}s>\n")
             idx += 1
 
             # Return block finished, we insert the return type if one was specified
-            if return_type is not None:
+            if tag in ["return", "yield"] and return_type is not None:
                 lines[idx - 1] += f"\n<{tag}type>{return_type}</{tag}type>\n"
+            elif raised_errors:
+                # rasied error
```
sgugger:
```suggestion
                # raised error
```

sgugger (approved):
Thanks!

PR #109, opened 2022-02-24
Fix documentation of --version in doc-builder build

### LINKED ISSUE #105: Documentation of --version
The current help text reads:

```
help="Version under which to push the files. Will not affect the actual files generated, as these are"
        " generated according to the `path_to_docs` argument.",
```

But now it will affect the files generated, in `kit/svelte.config.js`:

```
paths: {
			base:
				"/docs/" +
				(process.env.DOCS_LIBRARY || "transformers") +
				"/" +
				(process.env.DOCS_VERSION || "master") +
				"/" +
				(process.env.DOCS_LANGUAGE || "en")
		}
```

(it's used in the html files to reference the generated js files with an absolute path)

Maybe something like `DOCS_PATH` (= `path_to_docs`) should be passed instead? 

But the CI in transformers doesn't set `path_to_docs` :thinking: 

Edit: For the course we may want to reuse the `kit` folder of `doc-builder` with a fully configurable `paths.base` as well

sgugger:
This fixes the --help of the command `doc-builder build` for the `--version` argument.

Fixes #105

### DIFF
--- a/src/doc_builder/commands/build.py
+++ b/src/doc_builder/commands/build.py
@@ -176,8 +176,8 @@ def build_command_parser(subparsers=None):
     parser.add_argument(
         "--version",
         type=str,
-        help="Version under which to push the files. Will not affect the actual files generated, as these are"
-        " generated according to the `path_to_docs` argument.",
+        help="Version of the documentation to generate. Will default to the version of the package module (using "
+        "`master` for a version containing dev.",
     )
     parser.add_argument("--notebook_dir", type=str, help="Where to save the generated notebooks.", default=None)
     parser.add_argument("--html", action="store_true", help="Whether or not to build HTML files instead of MDX files.")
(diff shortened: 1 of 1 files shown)


src/doc_builder/commands/build.py
```diff
@@ -176,8 +176,8 @@ def build_command_parser(subparsers=None):
     parser.add_argument(
         "--version",
         type=str,
-        help="Version under which to push the files. Will not affect the actual files generated, as these are"
-        " generated according to the `path_to_docs` argument.",
+        help="Version of the documentation to generate. Will default to the version of the package module (using "
+        "`master` for a version containing dev.",
```
coyotte508, line 180:
```suggestion
        "`master` for a version containing dev).",
```

What does the package module refer to and "version containing dev". How does it get the default version (read the git tag, ...?)

sgugger: The module whose doc is being built. It has a `__version__` attribute containing its version.

As for version containing dev, it's the standard way of representing a version in development in Python, see for instance the [current version](https://github.com/huggingface/transformers/blob/7f921bcf47f2d6875feb6c1232dc6cfbeb4668e8/setup.py#L356) in Transformers.

coyotte508 (approved):
LGTM (with minor suggestion)

mishig25 (approved):
lgtm!

PR #110, opened 2022-02-24
Add CSS to generated HTML

coyotte508:
This also adds back the preview in dev mode

cc @Pierrci 

After this is merged, will have to regenerate past docs and open a PR similar to https://github.com/huggingface/doc-build/pull/2


kit/src/routes/__layout.svelte
```diff
@@ -0,0 +1,92 @@
+<script context="module" lang="ts">
+	import { base } from "$app/paths";
+
+	export async function load(input: LoadInput) {
+		if (prerendering || !import.meta.env.DEV) {
```
coyotte508, line 5: The issue with the previewing previously was the lack of `|| !import.meta.env.DEV`.

Because `prerendering` would be false on the client side, even if the html pages themselves were prerendered, as there was still some client side javascript

kit/svelte.config.js
```diff
@@ -22,6 +26,12 @@ const config = {
 			crawl: false // Do not throw if linked page doesn't exist (eg when forgetting the language prefix)
 		},
 
+		vite: {
+			build: {
+				sourcemap: Boolean(process.env.DOCS_SOURCEMAP)
```
coyotte508, line 31: Add optional source-map generation to debug problems =)

kit/src/routes/__layout.svelte
mishig25: great catch!

mishig25 (approved):
thanks a lot! I can update previous doc builds once it is merged

coyotte508:
> thanks a lot! I can update previous doc builds once it is merged

Thanks =)

I pushed the intermediate mdx files here: https://github.com/huggingface/doc-build/tree/intermediate-files, I manually found/replaced the svelte changes (adding `$lib` to the component imports & removing the `hydrate-props`).

Or you can regenerate them from scratch but it will replace the ColabDropdown w/ the new one. But it's probably fine too?

(Or I can do it if you want ^^)

mishig25:
@coyotte508 actually, could you regenerate past docs?
because of some specific dependency not being available, I can't generate `model_doc/qdqbert`

PR #111, opened 2022-02-24
Parse `yields` section from numpydocstring

mishig25:


src/doc_builder/utils.py
```diff
@@ -153,6 +153,11 @@ def start_section(docstring, section):
         for parameter in numydoc["Returns"]:
             returnttype = f":obj:`{parameter[1]}`" if parameter[1] else ""
             result += indent + ": ".join(s for s in [returnttype, stringify_arr(parameter[2])] if s)
+    if numydoc["Yields"]:
+        result = start_section(result, "Yields")
+        for parameter in numydoc["Yields"]:
+            returnttype = f":obj:`{parameter[1]}`" if parameter[1] else ""
```
sgugger, line 159:
```suggestion
            yield_type = f":obj:`{parameter[1]}`" if parameter[1] else ""
```

src/doc_builder/utils.py
```diff
@@ -153,6 +153,11 @@ def start_section(docstring, section):
...
+            result += indent + ": ".join(s for s in [returnttype, stringify_arr(parameter[2])] if s)
```
sgugger, line 160:
```suggestion
            result += indent + ": ".join(s for s in [yield_type, stringify_arr(parameter[2])] if s)
```

sgugger (approved):
Thanks for adding this!

PR #112, opened 2022-02-25
Clean up doctest artifacts

sgugger:
This PR cleans up the doctest artifacts that may appear in examples:
- flags to ignore outputs or skip
- a framework split variation since we can't use the regular one and it doesn't look like #63 is going to be merged any time soon.


mishig25 (approved):
lgtm!

I see that it is implement for `convert_md_to_mdx.py`. Do we not need it in `convert_rst_to_mdx.py` as well?

sgugger:
No, the doctests are only enabled o fully converted doc on our side :-)

PR #114, opened 2022-02-28
Log node errors

coyotte508:
For https://github.com/huggingface/transformers/pull/15710

### DIFF
--- a/src/doc_builder/commands/build.py
+++ b/src/doc_builder/commands/build.py
@@ -130,7 +130,7 @@ def build_command(args):
             print("Installing node dependencies")
             subprocess.run(
                 ["npm", "ci"],
-                stderr=subprocess.PIPE,
+                # stderr=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 check=True,
                 encoding="utf-8",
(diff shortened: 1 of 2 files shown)


mishig25 (approved):
lgtm! I think the output from node env is useful

src/doc_builder/commands/build.py
```diff
@@ -130,7 +130,7 @@ def build_command(args):
             print("Installing node dependencies")
             subprocess.run(
                 ["npm", "ci"],
-                stderr=subprocess.PIPE,
+                # stderr=subprocess.PIPE,
```
sgugger, line 133: If it's not used, why leave it commented out? Let's please clean up and remove it entirely.

src/doc_builder/commands/build.py
```diff
@@ -144,7 +144,7 @@ def build_command(args):
             print("Building HTML files. This will take a while :-)")
             subprocess.run(
                 ["npm", "run", "build"],
-                stderr=subprocess.PIPE,
+                # stderr=subprocess.PIPE,
```
sgugger, line 147: Same here.

src/doc_builder/commands/build.py
coyotte508: Done in 6adc10b5ede2281995ff04d95f3c94db2bbebc96

sgugger: Thanks!

PR #115, opened 2022-02-28
Make generated CSS empty

mishig25:
Do not output CSS file in sveltekit build to avoid problem mentioned [here](https://huggingface.slack.com/archives/C02GLJ5S0E9/p1646069793888439?thread_ts=1646055958.293249&cid=C02GLJ5S0E9).

There is still main body shifting to the left glitch, which is caused by [SubSideMenu and its breakpoints loading](https://github.com/huggingface/moon-landing/blob/4a16818bf77260ad9f3d23d530c990b859fac464/server/views/pages/DocPage/DocBuilderPage.svelte#L69). Will handle that problem in a separate moon-landing PR.

Rather than reverting the entire CSS addition as in [here](https://github.com/huggingface/doc-builder/pull/113), I'd like to keep it for dev purposes.

In the video below, I load the bert page 3 times for demo:

https://user-images.githubusercontent.com/11827707/156076438-d61b3aa3-becb-4580-b0d5-184878f56c7a.mov

### DIFF
--- a/kit/postbuild.sh
+++ b/kit/postbuild.sh
@@ -3,3 +3,7 @@ set -euo pipefail
 IFS=$'\n\t'
 
 cp src/routes/_toctree.yml build/_toctree.yml
+
+# To void conflict with Hub Tailwind CSS build, 
+# making doc-builder's PostCSS geenrated file an empty one
+> $(find . -regex '.*build.*.css')
--- a/kit/postbuild.sh
+++ b/kit/postbuild.sh
@@ -5,5 +5,8 @@ IFS=$'\n\t'
 cp src/routes/_toctree.yml build/_toctree.yml
 
 # To avoid conflict with Hub Tailwind CSS build, 
-# making doc-builder's PostCSS geenrated file an empty one
+# 1. making doc-builder's PostCSS geenrated file an empty one
+# 2. making `rel="stylesheet"` -> `rel="modulepreload"`
 > $(find . -regex '.*build.*.css')
+cd build
+find . -name '*.html' -exec sed -i '' 's/rel="stylesheet"/rel="modulepreload"/g' {} \;
--- a/kit/postbuild.sh
+++ b/kit/postbuild.sh
@@ -4,6 +4,6 @@ IFS=$'\n\t'
 
 cp src/routes/_toctree.yml build/_toctree.yml
 
-# To void conflict with Hub Tailwind CSS build, 
+# To avoid conflict with Hub Tailwind CSS build, 
 # making doc-builder's PostCSS geenrated file an empty one
 > $(find . -regex '.*build.*.css')


kit/postbuild.sh
```diff
@@ -3,3 +3,7 @@ set -euo pipefail
 IFS=$'\n\t'
 
 cp src/routes/_toctree.yml build/_toctree.yml
+
+# To void conflict with Hub Tailwind CSS build, 
+# making doc-builder's PostCSS geenrated file an empty one
+> $(find . -regex '.*build.*.css')
```
mishig25, line 9: technically, making the CSS file an empty file rather than deleting it to avoid `resouce not found errors`.

coyotte508 (approved):
Good work on diagnosing everything :muscle:

Pierrci:
> There is still main body shifting to the left glitch, which is caused by [SubSideMenu and its breakpoints loading](https://github.com/huggingface/moon-landing/blob/4a16818bf77260ad9f3d23d530c990b859fac464/server/views/pages/DocPage/DocBuilderPage.svelte#L69). Will handle that problem in a separate moon-landing PR.

If it's not linked to CSS or #110, it should be the same for stable versions and not just master no? Am I missing something? 🤔

mishig25:
@Pierrci I'm quite confused by this error.

In the video below, I've:
1. checked out `doc-builder` at https://github.com/huggingface/doc-builder/commit/c762adadd65ebaa36c6b8eda5d569b1eab310757, which is before https://github.com/huggingface/doc-builder/pull/110
2. build transformers master
3. toggle between `Disable cache`, as you can see, the glitch disappears, when `Disable cache` is clicked

https://user-images.githubusercontent.com/11827707/156082815-d995138d-b6cd-472f-a975-fb3258734f4d.mov

Pierrci:
@coyotte508 do you remember what version/commit of doc-builder you used to generate https://github.com/huggingface/doc-build/commit/5df5985cdcb5c3618c1ce9d2f4e8c1805717f9eb for example?

coyotte508:
The most recent one at the times, but I didn't use the python doc-builder, only the `kit` folder on the intermediate mdx files from this branch: https://github.com/huggingface/doc-build/tree/intermediate-files (not sure it's relevant)

As for the issue, maybe in `moon-landing` converting `rel="modulepreload"` in the generated HTML for the scripts to something else (iirc it makes the scripts blocking the render) / removing the `link rel="stylesheet"` would change how things are processed?

mishig25:
@coyotte508 you are right, removing stylesheet completely solves, the issue?
I can work on a postprocess python script that removes `<link rel="stylesheet" href="/docs/transformers/master/en/_app/assets/pages/__layout.svelte-xyz.css">` from every generated file.
Ideally, do you know if there is a setting in sveltekit or postcss so that CSS is not generated at all?

mishig25:
Steps to try:
1. Have moon-landing up to date
2. on doc-build, checkout branch empty_css_file_demo https://github.com/huggingface/doc-build/tree/empty_css_file_demo
3. go to this specific url http://localhost:5564/docs/transformers/master/en/model_doc/bert

https://github.com/huggingface/doc-builder/pull/115/commits/005fcf7fd4f3bffde32ebf411d5f3fb17b89cc79 solved the issue ?  because:
* with https://github.com/huggingface/doc-builder/pull/115/commits/005fcf7fd4f3bffde32ebf411d5f3fb17b89cc79, I have glitch on non-incognito Chrome (might be related to caching), do Not have glitch on incognito Chrome, Firefox
* without https://github.com/huggingface/doc-builder/pull/115/commits/005fcf7fd4f3bffde32ebf411d5f3fb17b89cc79, have glitches everywhere

coyotte508:
>  Ideally, do you know if there is a setting in sveltekit or postcss so that CSS is not generated at all?

I'm not sure, but there are plenty of options doing it otherwise. (including doing the replace inside `moon-landing` instead of `postbuild`).

mishig25:
> I'm not sure, but there are plenty of options doing it otherwise. (including doing the replace inside moon-landing instead of postbuild).

I've used post-build here rather than doing it on moon-landing to make the process less coupled, wdyt

kit/postbuild.sh
```diff
@@ -3,3 +3,10 @@ set -euo pipefail
 IFS=$'\n\t'
 
 cp src/routes/_toctree.yml build/_toctree.yml
+
+# To avoid conflict with Hub Tailwind CSS build, 
+# 1. making doc-builder's PostCSS geenrated file an empty one
+# 2. making `rel="stylesheet"` -> `rel="modulepreload"`
+> $(find . -regex '.*build.*.css')
+cd build
+find . -name '*.html' -exec sed -i '' 's/rel="stylesheet"/rel="modulepreload"/g' {} \;
```
coyotte508, line 12: Instead of replacing `stylesheet` by `modulepreload`, the `link` tag could be removed altogether

Pierrci: yes I agree, there's no interest in having it at all

PR #119, opened 2022-03-01
Support any package name in notebooks widget

sgugger:
This removes the hard-coded "transformers" to support other libraries (like datasets)


mishig25 (approved):
awesome!

PR #123, opened 2022-03-04
Rm numpydoc automatic conversion

mishig25:


sgugger (approved):
Thanks for cleaning this up!

PR #130, opened 2022-03-08
Implement framework specific content

mishig25:
# Framework specific content

## Input/syntax:
```html
<frameworkcontent>
<pytorch>
PYTORCH SPECIFIC CONTENT
</pytorch>
<tensorflow>
TENSORFLOW SPECIFIC CONTENT
</tensorflow>
<jax>
FLAX SPECIFIC CONTENT
</jax>
</frameworkcontent>
```

## Outputs/looks:

https://user-images.githubusercontent.com/11827707/157414522-a1fc1e03-17c3-4b80-ae27-fab58409f87f.mov

### How to test
1. Change `doc-build` repo's branch to `fwcontent` (which was built using this PR changes)
2. [Run this branch on moon-landing](https://github.com/huggingface/moon-landing/pull/2293)
3. Go to http://localhost:5564/docs/transformers/master/en/index

cc: @sgugger @stevhliu @LysandreJik 

merge with #63


kit/src/lib/FrameworkContentBlock.svelte
```diff
@@ -0,0 +1,77 @@
...
+    const localStorageKey = `hf_doc_framework_${framework}_is_hidden`;
+    const frameworkStore = getFrameworkStore(framework);
+
+    function toggleHidden(){
+        $frameworkStore = !$frameworkStore;
+        localStorage.setItem(localStorageKey, $frameworkStore ? "true" : "false");
+    }
+
+    onMount(() => {
+        if(localStorage.getItem(localStorageKey) === "true"){
```
mishig25, line 39: using [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) for persistance over multiple sessions for a user. wdy

kit/preprocess.js
```diff
@@ -117,6 +117,45 @@ export const docstringPreprocess = {
 	}
 };
 
+// Preprocessor that converts markdown into FrameworkContent
+// svelte component using mdsvexPreprocess
+export const frameorkcontentPreprocess = {
```
mishig25, line 122: using preprocessor to keep the syntax simpler
```html
<frameworkcontent>
<pytorch>
PYTORCH SPECIFIC CONTENT
</pytorch>
<tensorflow>
TENSORFLOW SPECIFIC CONTENT
</tensorflow>
<jax>
FLAX SPECIFIC CONTENT
</jax>
</frameworkcontent>
```

rather than letting a user write svelte with bunch of slots

kit/preprocess.js
```diff
@@ -117,6 +117,45 @@ export const docstringPreprocess = {
...
+		const REGEX_PYTORCH = /<pytorch>(((?!<pytorch>).)*)<\/pytorch>/ms;
+		const REGEX_TENSORFLOW = /<tensorflow>(((?!<tensorflow>).)*)<\/tensorflow>/ms;
+		const REGEX_JAX = /<jax>(((?!<jax>).)*)<\/jax>/ms;
+		const FRAMEWORKS = [
+			{framework: "pytorch", REGEX_FW: REGEX_PYTORCH, isExist: false},
+			{framework: "tensorflow", REGEX_FW: REGEX_TENSORFLOW, isExist: false},
+			{framework: "jax", REGEX_FW: REGEX_JAX, isExist: false},
+		];
+
+		content = await replaceAsync(content, REGEX_FRAMEWORKCONTENT, async (_, docstringBody) => {
```
coyotte508, line 134:
```suggestion
		content = await replaceAsync(content, REGEX_FRAMEWORKCONTENT, async (_, frameworkBody) => {
```

kit/preprocess.js
```diff
@@ -117,6 +117,45 @@ export const docstringPreprocess = {
...
+					<Markdown>
+					\n\n${fwContent}\n\n
+					</Markdown>
+					</svelte:fragment>`
+				}
+			}
+
+			const svleteProps = FRAMEWORKS.map(fw => `${fw.framework}={${fw.isExist}}`).join(" ");
+
+			return `<FrameworkContent ${svleteProps}>\n${svelteSlots}\n</FrameworkContent>`;
```
coyotte508, line 152:
```suggestion
			const svelteProps = FRAMEWORKS.map(fw => `${fw.framework}={${fw.isExist}}`).join(" ");

			return `<FrameworkContent ${svelteProps}>\n${svelteSlots}\n</FrameworkContent>`;
```

kit/preprocess.js
coyotte508: Maybe you could do that without a custom parser but with pure svelte code.

For example the `jax` component could be something like:

```ts
<FrameworkContentBlock framework="jax">
  <Markdown>
  	<slot />
  </Markdown>
</FrameworkContentBlock>
```

no?

kit/src/lib/FrameworkContentBlock.svelte
```diff
@@ -0,0 +1,77 @@
+<script lang="ts">
+    import type { SvelteComponent } from "svelte";
+    import type { Framework } from "./types";
+
+    import { onMount} from "svelte";
+    import { getFrameworkStore } from "./stores";
+	import IconPytorch from "./IconPytorch.svelte";
```
coyotte508, line 7: indent (also try using tabs instead of spaces, you can configure tab display width on github & in vs code)

kit/preprocess.js
mishig25: indeed, it would be possible to directly write svelte. There are 2 reasons why I prefer preprocessing:
1. `<frameworkcontent><jax>...</jax></frameworkcontent>` is simpler
2. Moreover, svelte/mdsvex compiler requires to have newlines so that a user can still write markdown inside slots. Thus `\n\n${fwContent}\n\n`, which is easy to do on preprocessor & hard to require if a user writes svelte directly
https://github.com/huggingface/doc-builder/blob/f3bb70cff4317f1feae4d0e134b5152cef7404c9/kit/preprocess.js#L143-L145

kit/preprocess.js
mishig25: handled in https://github.com/huggingface/doc-builder/pull/130/commits/4fc7d0f3d3a0ad56fb3b05c5484fcf7d33082590

kit/preprocess.js
coyotte508: What about this:

```html
<FrameworkContentBlock framework="jax">
  <Markdown>

<slot />

  </Markdown>
</FrameworkContentBlock>
```

But fair enough. Using svelte components would also require the user to order correctly the blocks, or add context logic with css ordering. And there is already existing code that preprocesses doc strings...

coyotte508 (approved):
Make sure to run prettier, to fix the spaces / tabs inside the svelte code

stevhliu:
Very nice, this looks so cool!

kit/src/lib/FrameworkContentBlock.svelte
Pierrci: looks good!

Pierrci (approved):
LGTM code-wise! 👍

julien-c (approved):
maybe rename `frameorkcontentPreprocess` to `frameworkcontentPreprocess`?

Other, let's get a final review from @gary149 and then deploy!

kit/src/lib/FrameworkContentBlock.svelte
```diff
@@ -0,0 +1,76 @@
...
+	function toggleHidden() {
+		$frameworkStore = !$frameworkStore;
+		localStorage.setItem(localStorageKey, $frameworkStore ? "true" : "false");
+	}
+
+	onMount(() => {
+		if (localStorage.getItem(localStorageKey) === "true") {
+			$frameworkStore = true;
+		}
+	});
```
julien-c, line 42: maybe it could be stored in DB (ie. only be a feature for logged in user) the same way we store dark mode (just an idea)

mishig25: on top of my head, not sure how we can pass this info to https://github.com/huggingface/moon-landing/blob/master/server%2Fviews%2Fpages%2FDocPage%2FDocBuilderPage.svelte#L54

cc: @coyotte508

Pierrci: one possibility is (maybe, not sure) to use svelte `context` for that, but rather convoluted - let's keep that for a second time :)

coyotte508: > one possibility is (maybe, not sure) to use svelte `context` for that, but rather convoluted

Would have never thought of that, but if the JS is compatible between the generated doc & the site maybe =o

But old generated version could break if svelte is upgraded.

I think maybe something with `window.postMessage` to trigger the doc -> moon-landing update and a `data-hide-x-framework` on `<body>` (or `<meta>` in `svelte:head`...) to get the initial value?

What about logged out users though? `localStorage` works better for them

> let's keep that for a second time :)

Definitely!

gary149:
On the UX we just did a call with Mishig, and he is going to improve some small details, we should be good when it's done. In a separate PR he will work and improving the navigation bar too (with logos associated with each framework block)

mishig25:
@gary149 updates:
1. Design-wise
a. smaller `Show/Hide` buttons & underline them on focus
b. Use negative `margin` rather than `absolute` & `translate`
2. Rightnvabar should be fully functionable (see attached video)
3. Framework specific content automatically opens if a user shares hashlink inside them

https://user-images.githubusercontent.com/11827707/157879973-c47d1ff3-724b-4b2a-ae22-7f7ec5e554bb.mov

gary149 (approved):

PR #134, opened 2022-03-10
Respect repo default branch

sgugger:
We hardcode `"master"` for the version in development but we should use the repo default branch (for instance `"main"`). This PR addresses that.

### DIFF
--- a/src/doc_builder/commands/build.py
+++ b/src/doc_builder/commands/build.py
@@ -66,6 +66,22 @@ def locate_kit_folder():
     # TODO for the future if asked for, if we can't locate the kit folder git clone doc-builder and cache it somewhere.
 
 
+def get_default_branch_name(repo_folder):
+    try:
+        p = subprocess.run(
+            ["git", "branch"],
+            stderr=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            check=True,
+            encoding="utf-8",
+            cwd=repo_folder,
+        )
+        branches = p.stdout.strip()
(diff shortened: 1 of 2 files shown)


src/doc_builder/commands/build.py
```diff
@@ -66,6 +66,22 @@ def locate_kit_folder():
...
+            stderr=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            check=True,
+            encoding="utf-8",
+            cwd=repo_folder,
+        )
+        branches = p.stdout.strip()
+        return "master" if "master" in branches else "main"
+    except Exception:
+        return "main"
```
mishig25, line 82: why is main returned on exception?

mishig25:
Update here to support main as well:
from:
```
if version == "master": 
```
to:
```
if version in ["master", "main"]: 
```
https://github.com/huggingface/doc-builder/blob/2e0f6e78eaf6fe3d02259769f23169bdf370bd4d/src/doc_builder/utils.py#L24-L30

mishig25 (approved):
Lgtm!

src/doc_builder/commands/build.py
LysandreJik: If you want to do it perfectly you could parse it from the result of `git symbolic-ref refs/remotes/origin/HEAD`, but it's fine for me to take `main` as the default.

LysandreJik (approved):
LGTM!

PR #135, opened 2022-03-10
Cleanup default branch check

sgugger:
#134 needed to be merged quickly. This PR takes the time to clean up things a bit and use a better check for the main branch name.


LysandreJik (approved):
This looks super good to me!

PR #136, opened 2022-03-10
Don't replace return type

sgugger:
If a return statement takes several lines (like models in the Transformers library) the `return_type` is replaced by None at the second line of that return statement. This PR fixes that.


LysandreJik (approved):
Yes, didn't try it locally but it looks good to me!

PR #137, opened 2022-03-10
Add integration test for doc building

sgugger:
This PR adds a new GitHub action that builds the doc of Accelerate on each PR. This makes a nice integration test that stays short.


LysandreJik (approved):
This is great, thanks for implementing it!

PR #138, opened 2022-03-11
Adds templates for doc building workflows

sgugger:
This PR adds templates for the jobs that:
- build the doc at each commit on master
- build the doc for each PR
- clean the doc after each PR

The last two have been tested on Accelerate. The first one still needs testing.

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -20,6 +20,8 @@ on:
 jobs:
   build_main_documentation:
     runs-on: ubuntu-latest
+    container:
+      huggingface/transformers-doc-builder
 
     steps:
       - uses: actions/checkout@v2
@@ -61,8 +63,6 @@ jobs:
 
       - name: Setup environment
         run: |
-          pip install -U click # Sometimes we end up with click 7.0 and we need click 8.0 for black
-
           pip uninstall -y doc-builder
           cd doc-builder
           git pull origin main
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -6,13 +6,12 @@ on:
       commit_sha:
         required: true
         type: string
-      pr_number:
-        required: true
-        type: string
       package:
         required: true
         type: string
       path_to_doc:
+        type: string
+      notebook_folder:
         type: string
     secrets:
       token:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_pr_documentation.yml
```diff
@@ -0,0 +1,119 @@
...
+        type: string
+      pr_number:
+        required: true
+        type: string
+      package:
+        required: true
+        type: string
+      path_to_doc:
+        type: string
+
```
mishig25, line 17: in [transformers pr_doc](https://github.com/huggingface/transformers/blob/master/.github/workflows/build_dev_documentation.yml#L6-L8), there are lines about concurrency

```py
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

do we need them here as well?

.github/workflows/build_main_documentation.yml
```diff
@@ -0,0 +1,137 @@
...
+        type: string
+      path_to_doc:
+        type: string
+      notebook_folder:
+        type: string
+    secrets:
+      token:
+        required: true
+
+jobs:
```
mishig25, line 20: in transformers [build doc workkflow](https://github.com/huggingface/transformers/blob/master/.github/workflows/build_documentation.yml#L45-L53), there is a `job` about `cache`:
```yml
      - name: Loading cache.
        uses: actions/cache@v2
        id: cache
        with:
          path: ~/.cache/pip
          key: v1-test_build_doc
          restore-keys: |
            v1-test_build_doc-${{ hashFiles('setup.py') }}
            v1-test_build_doc
```

do we need them here as well? or add them as a `workflow.inputs`

sgugger: No, this is using the docker image now.

.github/workflows/build_pr_documentation.yml
sgugger: This should be on the workflows that apply this template (if the library wants it), see for instance in [this PR](https://github.com/huggingface/doc-builder/pull/138) (GitHub doesn't let me link to a specific line but it's line 6 in the build_pr_documentation file)

mishig25 (approved):

.github/workflows/build_main_documentation.yml
```diff
@@ -0,0 +1,137 @@
...
+      package:
+        required: true
+        type: string
+      path_to_doc:
+        type: string
+      notebook_folder:
+        type: string
+    secrets:
+      token:
+        required: true
```
LysandreJik, line 18: Very clean

LysandreJik (approved):
Looks great, thanks for spending some time on this!

PR #139, opened 2022-03-14
Implement preview command

mishig25:
# Dev command with hot reloading

`doc-builder preview datasets ~/Desktop/datasets/docs/source`

I think this will make the experience writing docs better 🙂

https://user-images.githubusercontent.com/11827707/158227378-1bb04d04-4010-48b3-b039-9d2bdc72c4cf.mov


src/doc_builder/build_doc.py
```diff
@@ -213,7 +221,7 @@ def build_mdx_files(package, doc_folder, output_dir, page_info):
             "The deployment of the documentation will fail because of the following errors:\n" + "\n".join(all_errors)
         )
 
-    return anchor_mapping
+    return anchor_mapping, src_py_path_mapping
```
mishig25, line 224: `src_py_path_mapping` is used to know wich `mdx` file to updated when a docstring is updated inside `python` file

src/doc_builder/commands/dev.py
```diff
@@ -0,0 +1,218 @@
...
+from datetime import datetime, timedelta
+from pathlib import Path
+from threading import Thread
+
+from doc_builder import build_doc
+from doc_builder.commands.build import check_node_is_available, locate_kit_folder
+from doc_builder.commands.convert_doc_file import find_root_git
+from doc_builder.utils import read_doc_config
+from watchdog.events import FileSystemEventHandler
+from watchdog.observers import Observer
```
sgugger, line 32: This is a new dependency, so would prefer to have this as a soft dependency. This means those imports and the class below should be moved under an `if is_watchdog_available:` test.

For the value of `is_watchdog_available` we should use a `try / except` block to see if we can import this or not. Then the `dev_command` would start with a test raising an `ImportError` if `is_watchdog_available` is `False`.

This would solve the doc building issue in the CI.

src/doc_builder/autodoc.py
```diff
@@ -278,6 +278,19 @@ def get_source_link(obj, page_info):
...
 
 
+def get_source_path(object_name, package):
+    """
+    Find a path to file in which given object was defined.
+
+    Args:
+    - **object_name** (`str`) -- The name of the object to retrieve.
+    -- **package** (`types.ModuleType`) -- The package to look into.
+    """
```
sgugger, line 288:
```suggestion
    Args:
    - object_name (`str`): The name of the object to retrieve.
    - package (`types.ModuleType`): The package to look into.
    """
```

src/doc_builder/build_doc.py
```diff
@@ -153,6 +156,7 @@ def build_mdx_files(package, doc_folder, output_dir, page_info):
     output_dir = Path(output_dir)
     os.makedirs(output_dir, exist_ok=True)
     anchor_mapping = {}
+    src_py_path_mapping = {}
```
sgugger, line 159:
```suggestion
    source_files_mapping = {}
```
Same length but more understandable.

sgugger (commented):
Looks like a great addition, thanks for working on this! I have a few comments on the variable names and we need to handle `watchdog` as a soft dependency (that the user should install only if they want to use `doc-builder dev`).

The name `dev` is not super intuitive to me, is it something that comes from node? I'd prefer something like `doc-builder preview` (or if you have another idea I'm all ears!).

src/doc_builder/build_doc.py
mishig25: handled in https://github.com/huggingface/doc-builder/pull/139/commits/31c4207a7069d765dc2b633d33d698e1e9da986b

src/doc_builder/commands/dev.py
mishig25: handled in https://github.com/huggingface/doc-builder/pull/139/commits/317030a82700f2243bf3d55eb744ca4c8b632419

src/doc_builder/commands/dev.py
```diff
@@ -0,0 +1,218 @@
...
+                shutil.copytree(f, dest)
+            else:
+                shutil.copy(f, dest)
+
+        # Node
+        env = os.environ.copy()
+        env["DOCS_LIBRARY"] = args.library_name
+        env["DOCS_VERSION"] = args.version
+        env["DOCS_LANGUAGE"] = args.language
+        Thread(target=start_sveltekit_dev, args=(tmp_dir, env, args)).start()
```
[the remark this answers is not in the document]
mishig25: this thread should be running as long as the program is running

mishig25:
@sgugger thanks a lot for the suggestions!
1. Renamed command `dev` -> `preview`
2. Added soft dependency `watchdog`
3. Renamed variables as suggested

src/doc_builder/commands/preview.py
```diff
@@ -0,0 +1,225 @@
...
+        ["npm", "run", "dev"],
+        check=True,
+        encoding="utf-8",
+        cwd=working_dir,
+        env=env,
+    )
+
+
+def preview_command(args):
+    if is_watchdog_available():
```
sgugger, line 144: To avoid everything being indented here, I'd test the negation and raise the `ImportError`, then just unindent the whole function.

sgugger (approved):
Thanks for adding all of that! I just left one last styling suggestion.

For the `Thread`, are we sure that exiting will close it (it's not the case with `Process` if I remember correctly, so just want to be 100% sure)

src/doc_builder/commands/preview.py
```diff
@@ -175,9 +175,9 @@ def preview_command(args):
             kit_routes_folder = tmp_dir / "kit" / "src" / "routes"
             # files/folders cannot have a name that starts with `__` since it is a reserved Sveltekit keyword
             for dir in output_path.glob("**/__*/*"):
-                dir.rmdir()
+                subprocess.run(["rm", "-rf", str(dir)])
```
sgugger, line 178:
```suggestion
                shutil.rmtree(dir)
```
This is cleaner.

src/doc_builder/commands/preview.py
```diff
@@ -175,9 +175,9 @@ def preview_command(args):
...
             for f in output_path.glob("**/__*"):
-                f.unlink()
+                subprocess.run(["rm", "-rf", str(f)])
```
sgugger, line 180:
```suggestion
                os.unlink(f)
```

src/doc_builder/commands/preview.py
sgugger: It only works for dir, so you should make a test. It's still better than using `subprocess` for this.

mishig25: handled in https://github.com/huggingface/doc-builder/pull/139/commits/635938b0ec5030013266dc2d8299d4743056da52

src/doc_builder/commands/preview.py
mishig25: https://github.com/huggingface/doc-builder/pull/139/commits/979e8d981a887d911374d9d5e94fdf2b1e22a0ad

mishig25:
@LysandreJik please feel free to retry the preview command.
You may need to reinstall the CLI if you face some error starting it up.

```
doc-builder preview datasets ~/Desktop/datasets/docs/source/
```

Fixed the relative path issue (should be working whether you provide absolute or relative path to docs folder)

LysandreJik:
Thanks for the ping, will try it again.

PR #141, opened 2022-03-15
Frontend for `yields` & `raises` docstrings

mishig25:
closes https://github.com/huggingface/doc-builder/issues/172

Screenshot 2022-03-15 at 17 51 44
 
Screenshot 2022-03-15 at 17 58 23


kit/preprocess.js
```diff
@@ -87,6 +91,30 @@ export const docstringPreprocess = {
...
+				const raiseDesc = docstringBody.match(REGEX_RAISEDESC)[1];
+				const { code } = await mdsvexPreprocess.markup({ content: raiseDesc, filename });
+				svelteComponent += ` raiseDescription={${JSON.stringify(code)}} `;
+			}
+
+			if (docstringBody.match(REGEX_RAISETYPE)) {
+				const raiseType = docstringBody.match(REGEX_RAISETYPE)[1];
+				const { code } = await mdsvexPreprocess.markup({ content: raiseType, filename });
+				svelteComponent += ` raiseType={${JSON.stringify(code)}} `;
+			}
```
coyotte508, line 116: Eventually, something like a key/value map for the properties and generating the component string at once seems cleaner (rather than multiple `svellteComponent += ...` followed by a closing tag)

coyotte508 (approved):
LGTM

sgugger:
I know @LysandreJik was using warning blocks for the raises block to have something in color. Not sure whether it's something we want or not.

LysandreJik:
I like the current approach! Thanks @mishig25!

mishig25:
@LysandreJik 
> I like the current approach! Thanks @mishig25!

do you mean: you prefer using <Tip warning> for the raise?
if so, should I make this PR only render `yields`, which should look like `returns`?

LysandreJik:
Sorry, just realized I didn't answer: I was mentioning I like the approach you have taken in your PR, @mishig25

LysandreJik:
Bump on this, we're using `Raises` quite a bit in `huggingface_hub` and we would benefit from having this in :)

cc @Wauplin

PR #142, opened 2022-03-15
Git clone for kit

sgugger:
The kit subfolder can't be packaged with `doc-builder` as it makes a tar file that is too heavy (105MB) and PyPi refuses it. This PR removes it from the package (deletion of the manifest file) and introduces a mechanism for cloning and caching the repo to access the kit subfolder.

### DIFF
--- a/src/doc_builder/utils.py
+++ b/src/doc_builder/utils.py
@@ -158,7 +158,6 @@ def get_cached_repo():
         _ = subprocess.run(
             "git clone https://github.com/huggingface/doc-builder.git".split(),
             stderr=subprocess.PIPE,
-            stdout=subprocess.PIPE,
             check=True,
             encoding="utf-8",
             cwd=DOC_BUILDER_CACHE,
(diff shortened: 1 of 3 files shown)


src/doc_builder/utils.py
```diff
@@ -97,3 +106,71 @@ def get_doc_config():
...
+    cache_repo_path = Path(DOC_BUILDER_CACHE) / "doc-builder-repo"
+    if not cache_repo_path.is_dir():
+        print(
+            "To build the HTML doc, we need the kit subfolder of the `doc-builder` repo. Cloning it and caching at "
+            f"{cache_repo_path}."
+        )
+        _ = subprocess.run(
+            "git clone https://github.com/huggingface/doc-builder.git".split(),
+            stderr=subprocess.PIPE,
+            stdout=subprocess.PIPE,
```
LysandreJik, line 161: These line will prevent logs from being output to the command line, won't logs be useful in some situations?

LysandreJik (approved):
Looks good, left a comment regarding the `subprocess.run`.

sgugger:
Merging this, I'll resolve the packaging issues in a separate PR (with or without a MANIFEST file).

PR #147, opened 2022-03-16
Use the utils `locate_kit_folder`

sgugger:
A final cleanup following up from #142: actually use the new `locate_kit_folder` funciton :sweat_smile:


mishig25 (approved):
Lgtm!

PR #149, opened 2022-03-17
Optimum fixes

sgugger:
In a namespace module (like `optimum`), submodules won't directly appear as attributes. You need to import them at least one explicitly before they do. This PR does that.


LysandreJik (approved):
Looks good to me, thanks for investigating @sgugger @lewtun!

PR #151, opened 2022-03-24
Switch default branch to main and fix tests

sgugger:
This PR changes the default branch to "main" everywhere master is used, in particular for the notebook building job (for the rest it's the default in internal functions, which is trumped by the version set when building the doc anyway).

It also fixes a few tests due to the recent reorganization in Transformers splitting `file_utils` in several submodules.


mishig25 (approved):
thanks for this! does it mean that for `datasets` need to specifically supply version `master` since they have not migrated to `main` yet?

sgugger:
No, the command picks the default branch in the `_config.py` and sets it to `master`.

LysandreJik (approved):
LGTM!

PR #152, opened 2022-03-25
`main_branch` input for `build_main_documentation.yml`

mishig25:


sgugger:
No as said before, the name for the default branch is already customizable with the doc config. This would also only build the documentation for the main version and not the released versions.

mishig25:
oh indeed. forgot about that solution 😊

PR #153, opened 2022-03-25
Gh workflows support `--not_python_module` flag

mishig25:
so that we can use GH templates in course new frontend Ci [here](https://github.com/huggingface/course/pull/35#issuecomment-1078128606)

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -13,6 +13,10 @@ on:
         type: string
       notebook_folder:
         type: string
+      # supply --not_python_module for HF Course
+      module_flag:
+        type: string
+        default: ''
     secrets:
       token:
         required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/build_main_documentation.yml
```diff
@@ -13,6 +13,10 @@ on:
         type: string
       notebook_folder:
         type: string
+      # supply --not_python_module for HF Course
+      module_flag:
```
sgugger, line 17: I'd call this `additional_args` so we can use it to pass anything.

.github/workflows/build_main_documentation.yml
```diff
@@ -13,6 +13,10 @@ on:
...
+        type: string
+        default: ''
```
sgugger, line 19: The default will be this, so this line can be removed.
```suggestion
```

.github/workflows/build_pr_documentation.yml
```diff
@@ -14,6 +14,10 @@ on:
         type: string
       path_to_doc:
         type: string
+      # supply --not_python_module for HF Course
+      module_flag:
+        type: string
+        default: ''
```
sgugger, line 20: Same comments on this file.

sgugger (commented):
Thanks for adding this! I think we can make the new arg more general.

.github/workflows/build_main_documentation.yml
mishig25: I see, https://github.com/huggingface/doc-builder/pull/153/commits/35a58fa1ee0fd6776e32f2519ec83dcda72a760d

LysandreJik (approved):
This is a great addition, thanks @mishig25!

mishig25:
Thanks a lot fir the reviews 😊 merging it since trying to get the doc-build asap

mishig25:
Thanks a lot fir the reviews 😊 merging it since trying to get the doc-build asap

PR #154, opened 2022-03-25
Skip package install if --not_python_module is provided

lewtun:
This PR adds logic to skip the `pip install .[dev]` step in the doc building templates if the `--not_python_module` argument is provided. This should solve the failing build for the course [here](https://github.com/huggingface/course/runs/5690210360?check_suite_focus=true)

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -72,7 +72,7 @@ jobs:
           pip install .
           cd ..
 
-          if [ "${{ inputs.additional_args }}" != "--not_python_module" ]
+          if [ "${{ inputs.additional_args }}" != *"--not_python_module"* ]
           then
             cd ${{ inputs.package }}
             pip install .[dev]
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -72,7 +72,7 @@ jobs:
           pip install .
           cd ..
 
-          if [ "${{ inputs.additional_args }}" != *"--not_python_module"* ]
+          if [[ "${{ inputs.additional_args }}" != *"--not_python_module"* ]];
           then
             cd ${{ inputs.package }}
             pip install .[dev]
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -9,7 +9,7 @@ on:
       package:
         required: true
         type: string
-      path_to_doc:
+      path_to_docs:
         type: string
       notebook_folder:
         type: string
@@ -72,8 +72,11 @@ jobs:
           pip install .
           cd ..
 
-          cd ${{ inputs.package }}
-          pip install .[dev]
+          if [ "${{ inputs.additional_args }}" != "--not_python_module" ]
+          then
+            cd ${{ inputs.package }}
+            pip install .[dev]
+          fi
           cd ..
 
       - name: Setup git
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -63,7 +63,7 @@ jobs:
           pip install .
           cd ..
 
-          if [ "${{ inputs.additional_args }}" != *"--not_python_module"* ]
+          if [[ "${{ inputs.additional_args }}" != *"--not_python_module"* ]];
           then
             cd ${{ inputs.package }}
             pip install .[dev]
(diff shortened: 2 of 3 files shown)


.github/workflows/build_main_documentation.yml
```diff
@@ -9,7 +9,7 @@ on:
       package:
         required: true
         type: string
-      path_to_doc:
+      path_to_docs:
```
lewtun, line 12: I think there was a typo in the original definition, so I've updated this to reflect the variable name in the rest of the YAML

.github/workflows/build_pr_documentation.yml
```diff
@@ -63,8 +63,11 @@ jobs:
           pip install .
           cd ..
 
-          cd ${{ inputs.package }}
-          pip install .[dev]
+          if [ "${{ inputs.additional_args }}" != "--not_python_module" ]
```
mishig25, line 66: `inputs.additional_args` can potentially hold more args like `--not_python_module --another_arg1 --another_arg2`

would it be possible to check something like:
`if inputs.additional_args includes --not_python_module`

lewtun: Nice idea! Will add that :)

.github/workflows/build_main_documentation.yml
```diff
@@ -72,8 +72,11 @@ jobs:
           pip install .
           cd ..
 
-          cd ${{ inputs.package }}
-          pip install .[dev]
+          if [ "${{ inputs.additional_args }}" != *"--not_python_module"* ]
```
lewtun, line 75: This checks that `--not_python_submodule` is not a substring of `inputs.additional_args`

sgugger (approved):
LGTM, as long as there is the suggestion of mishig added.

.github/workflows/build_main_documentation.yml
lewtun: Ah, it turns out you need to use double brackets with the wildcard outside the string: see [here](https://linuxize.com/post/how-to-check-if-string-contains-substring-in-bash/#using-wildcardshttps://linuxize.com/post/how-to-check-if-string-contains-substring-in-bash/#using-wildcards) and [here](https://stackoverflow.com/questions/229551/how-to-check-if-a-string-contains-a-substring-in-bash). I've fixed this in https://github.com/huggingface/doc-builder/pull/154/commits/2eb9af4659112886f9e52fcad439e2461771d365

sgugger: Just to be sure, can you run a test on this branch of the doc-builder to build the course doc before merging?

lewtun:
I see that #151 will fix the failing tests - should I wait until that PR is merged before merging this one?

sgugger:
The tests are irrelevant to your PR, so you can ignore. Just make sure you try this workflow from your branch on the course repo once before merging, to make sure all is well.

.github/workflows/build_main_documentation.yml
lewtun: Sure, good idea!

src/doc_builder/commands/build.py
```diff
@@ -62,7 +62,7 @@ def build_command(args):
                 "the doc-builder package installed, so you need to run the command from inside the doc-builder repo."
             )
 
-    if args.not_python_module:
+    if args.not_python_module and args.version is None:
```
lewtun, line 65: I found that we need to check that `args.version` isn't provided because the PR build uses this flag to define the subfolders. Without this change, the docs are always built to `course/main/` and we get a 

```
mv: cannot stat '../build_dir/course/pr_35': No such file or directory
```

error when we try to push to the `doc-build-dev` repo. Running the course CI again to double-check

mishig25: oh indeed :) thanks for catching this

mishig25:
lgtm from my side!

lewtun:
The course docs built successfully with this branch: https://github.com/huggingface/doc-build-dev/tree/main/course/pr_35/en

FYI I found that I had to manually create a `course` directory in the `doc-build-dev` [repo](https://github.com/huggingface/doc-build-dev) because the CI was throwing the following error:

```
mv: cannot move '../build_dir/course/pr_35' to 'course/pr_35': No such file or directory
```

I'm not sure if this was an isolated event, but happy to add a note to the README if you think it's needed. 

Merging this to wrap up the course build :)

PR #157, opened 2022-03-25
Fixes for the doc-builder style command

sgugger:
This PR introduces two small fixes to make the doc-builder style command work on Transformers (it needs to access the doc-builder config as we have some specific patterns to deal with in the generic docstrings)


LysandreJik (approved):
LGTM!

PR #158, opened 2022-03-25
Fix `build` command in README.md

lhoestq:
The `build` keyword is missing


sgugger (approved):

sgugger:
Thanks for the fix!

PS: Please ping someone for review when creating a PR or it will probably be ignored ;-)

PR #159, opened 2022-03-25
Disable notebook building for non-main builds

sgugger:
This PR removes the notebook build when the job is called with a released version and not the main/master version: the doc building job always pushes notebooks to the main branch of the notebooks repo, so it shouldn't be updated for a released version.


LysandreJik (approved):
Sounds good!

Issue #160, opened 2022-03-28
Library package not installed while building docs


albertvillanova:
We are having issues with documentation building:
- https://github.com/huggingface/datasets/runs/5716300989?check_suite_focus=true

```
ValueError: There was an error when converting ../datasets/docs/source/package_reference/main_classes.mdx to the MDX format.
Unable to find datasets.filesystems.S3FileSystem in datasets. Make sure the path to that object is correct.
```

After some investigation: 
- the issues started just after this PR was merged:
  - #154 
- looking at the corresponding action summary, it seems that the `datasets[dev]` package installation was skipped

CC: @lewtun

lewtun:
Oh indeed, I see a weird parse error:

```
/__w/_temp/75bf1d4b-4452-4c1d-8927-838e5094620a.sh: 7: [[: not found
```

I'll revert my PR and investigate how to fix this

albertvillanova:
@lewtun please note that I'm just guessing... Maybe your PR is not the source of the issue...

lewtun:
I think you're guess is right @albertvillanova - from this [SO answer](https://stackoverflow.com/questions/12230690/string-comparison-in-bash-not-found) it seems that the lack of bash builtins is causing the problem. I've reverted the commit in #161 so we can now see if that resolves the issue

albertvillanova:
@lewtun I confirm that the docs building works as expected after your reversion:
- https://github.com/huggingface/datasets/runs/5718022417?check_suite_focus=true

PR #162, opened 2022-03-28
Skip package installation if `--not_python_module` provided

lewtun:
This PR reverts the revert in #161 to properly set the shell to `bash` when we run the `Setup environment` step. As far as I can tell, `ubuntu-latest` actually runs `dash` by default, so it couldn't parse the double `[[` brackets that we use for substring matching, e.g. in https://github.com/huggingface/doc-builder/issues/160 we encounter the following error in the CI:

```
/__w/_temp/75bf1d4b-4452-4c1d-8927-838e5094620a.sh: 7: [[: not found
```

Setting `shell: bash` in the CI step resolves the issue and here are two successful builds of the docs on `datasets` and `course` based on this branch:

* https://github.com/huggingface/datasets/runs/5718880095?check_suite_focus=true
* https://github.com/huggingface/course/runs/5719043810?check_suite_focus=true (the add doc comment step fails, but that's because we need to deploy `moon-landing` with some updates)

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -65,6 +65,7 @@ jobs:
           fi
 
       - name: Setup environment
+        shell: bash
         run: |
           pip uninstall -y doc-builder
           cd doc-builder
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -9,7 +9,7 @@ on:
       package:
         required: true
         type: string
-      path_to_doc:
+      path_to_docs:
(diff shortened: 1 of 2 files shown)


.github/workflows/build_main_documentation.yml
```diff
@@ -65,15 +65,19 @@ jobs:
           fi
 
       - name: Setup environment
+        shell: bash
```
mishig25, line 68: thanks for the catch!

mishig25 (approved):

sgugger (approved):

PR #163, opened 2022-03-28
Support anchor names for inherited methods

sgugger:
There is a problem in the way anchors are generated for methods of classes that are inherited from a superclass. The problem was first reported in [this PR](https://github.com/huggingface/datasets/pull/3999) but it happens more broadly when you have an object (let's say `BertTokenizer`) and a method (let's say the `__call__` magic method) that you are asking in the documentation, but that method's documentation is actually written in the superclass (in our example `PreTrainedTokenizerBase`).

Currently, doc-builder will create the documentation of this method under the anchor `PreTrainedTokenizerBase.__call__` which has two problems:
- if we were in a page with two different tokenizers documenting that method, we would get the same anchor twice
- more annoying is that this method can never be linked to via inter-links. When asking for a link to this method via ``[`BertTokenizer.__call__`]``, we get a link to the `BertTokenizer.__call__` anchor

And even more insidious, the object found by `doc-builder` when looking at `BertTokenizer.__call__` is `PreTrainedTokenizerBase.__call__` (that's why this is the generated anchor name), so when resolving ``[`BertTokenizer.__call__`]``, it tries to find in the anchors mapping `PreTrainedTokenizerBase.__call__`, but in that mapping, the method is under the child's name `BertTokenizer.__call__`, so the reference is not resovled.

This PR solves those issues the following way:
- it adds an `anchor_name` to the `document_object` method, to force the anchor name to be `{documented_class_anchor}.{method}` in those situations
- if the `anchor_name` is different from the object name (as above) it saves both names in the anchor mapping
- when it's time to build links from those mappings, it creates, for the object name a link to the proper anchor name.

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -300,7 +300,7 @@ def get_source_path(object_name, package):
     return obj_path
 
 
-def document_object(object_name, package, page_info, full_name=True):
+def document_object(object_name, package, page_info, full_name=True, anchor_name=None):
     """
     Writes the document of a function, class or method.
 
@@ -308,6 +308,7 @@ def document_object(object_name, package, page_info, full_name=True):
         object_name (`str`): The name of the object to document.
         package (`types.ModuleType`): The package of the object.
         full_name (`bool`, *optional*, defaults to `True`): Whether to write the full name of the object or not.
+        anchor_name (`str`, *optional*): The name to give to the anchor for this object.
     """
     if page_info is None:
         page_info = {}
@@ -322,7 +323,8 @@ def document_object(object_name, package, page_info, full_name=True):
         # Propreties have no __module__ or __name__ attributes, but their getter function does.
         obj = obj.fget
 
-    anchor_name = get_shortest_path(obj, package)
+    if anchor_name is None:
+        anchor_name = get_shortest_path(obj, package)
     if full_name and anchor_name is not None:
         name = anchor_name
     else:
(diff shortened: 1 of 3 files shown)


src/doc_builder/autodoc.py
```diff
@@ -414,14 +416,25 @@ def autodoc(object_name, package, methods=None, return_anchors=False, page_info=
...
+                anchor_name=anchor_name,
             )
             if check is not None:
                 errors.append(check)
             documentation += '\n<div class="docstring">' + method_doc + "</div>"
             if return_anchors:
-                anchors.append(f"{anchors[0]}.{method}")
+                # The anchor name of the method might be different from its
+                method = find_object_in_package(f"{anchors[0]}.{method}", package=package)
+                method_name = get_shortest_path(method, package=package)
```
sgugger, line 433: This replicates what the doc-builder will do when we try to find we try to resolve link to ``[`BertTokenizer.__call__`]`` -> it will get a different `method_name` compared to the `anchor_name`.

src/doc_builder/autodoc.py
```diff
@@ -463,7 +476,10 @@ def _resolve_link(search):
         if anchor not in mapping:
             return f"`{object_name}`"
         page = f"{prefix}{mapping[anchor]}"
-        return f"[{object_name}]({page}#{anchor}){last_char}"
+        if "#" in page:
+            return f"[{object_name}]({page}){last_char}"
+        else:
+            return f"[{object_name}]({page}#{anchor}){last_char}"
```
sgugger, line 482: Some of the pages already have the anchor in the mapping (the ones where the object name is different from the anchor name).

src/doc_builder/build_doc.py
```diff
@@ -211,7 +212,11 @@ def build_mdx_files(package, doc_folder, output_dir, page_info):
 
         if new_anchors is not None:
             page_name = str(file.with_suffix("").relative_to(doc_folder))
-            anchor_mapping.update({anchor: page_name for anchor in new_anchors})
+            for anchor in new_anchors:
+                if isinstance(anchor, tuple):
+                    anchor_mapping.update({a: f"{page_name}#{anchor[0]}" for a in anchor[1:]})
```
sgugger, line 217: When we have several anchors for the same object, the first one is the real anchor name, the other ones, aliases we register.

LysandreJik: Sounds good!

LysandreJik (approved):
Looks good to me!

sgugger:
Merging this without waiting for Mishig's approval since he is on vacation, but will address any comments he has in a follow-up PR if needed :-)

PR #164, opened 2022-03-29
Add multilingual support for CI templates

lewtun:
This PR adds multilingual support for the CI templates via a new `languages` argument. Thanks to a tip from @sgugger, the current logic is _opt-in_ which means that multilingual docs are only created if the `languages` argument is provided and the source docs are structured as follows:

```
doc_folder
├── en
│   ├── _toctree.yml
    ...
└── es
    ├── _toctree.yml
    ...
```

An [example](https://github.com/huggingface/doc-build-dev/tree/main/course/pr_39) showing a successful build for the course can be found in this PR: https://github.com/huggingface/course/pull/39

Similarly, an example showing that the current implementation works for monolingual docs can be found here: https://github.com/huggingface/datasets/pull/4055

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -16,6 +16,10 @@ on:
       # supply --not_python_module for HF Course
       additional_args:
         type: string
+      languages:
+      # supply space-separated language codes
+        type: string
+        default: 'en'
     secrets:
       token:
         required: true
@@ -95,6 +99,7 @@ jobs:
           cp doc-build/${{ inputs.package }}/_versions.yml build_dir/${{ inputs.package }}
 
       - name: Make documentation
+        shell: bash
         env:
           NODE_OPTIONS: --max-old-space-size=6656
         run: |
@@ -102,8 +107,14 @@ jobs:
 
           if [ -z "${{ inputs.notebook_folder }}" ]
           then
-            cd doc-builder &&
-            doc-builder build ${{ inputs.package }} ../${{ env.doc_folder }} --build_dir ../build_dir --clean --html ${{ inputs.additional_args }} &&
+            IFS=', ' read -r -a langs <<< "${{ inputs.languages }}"
+
+            cd doc-builder
+            for lang in "${langs[@]}"
+            do
+                echo "Generating docs for language $lang"
+                doc-builder build ${{ inputs.package }} ../${{ env.doc_folder }}/$lang --build_dir ../build_dir --clean --html ${{ inputs.additional_args }} --language $lang
+            done
             cd ..
           else
             cd doc-builder &&
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -104,30 +104,28 @@ jobs:
           NODE_OPTIONS: --max-old-space-size=6656
         run: |
           echo "doc_folder has been set to ${{ env.doc_folder }}"
+          cd doc-builder
+          args="--build_dir ../build_dir --clean --html ${{ inputs.additional_args }}"
 
-          if [ -z "${{ inputs.notebook_folder }}" ];
+          if [ ! -z "${{ inputs.notebook_folder }}" ];
           then
-            IFS=', ' read -r -a langs <<< "${{ inputs.languages }}"
-
-            cd doc-builder
-            if [ -z "${{ inputs.languages }}" ];
-            then
-              echo "languages not provided, defaulting to English"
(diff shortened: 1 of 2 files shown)


.github/workflows/build_main_documentation.yml
```diff
@@ -95,15 +99,22 @@ jobs:
...
           then
-            cd doc-builder &&
-            doc-builder build ${{ inputs.package }} ../${{ env.doc_folder }} --build_dir ../build_dir --clean --html ${{ inputs.additional_args }} &&
+            IFS=', ' read -r -a langs <<< "${{ inputs.languages }}"
+
+            cd doc-builder
+            for lang in "${langs[@]}"
+            do
+                echo "Generating docs for language $lang"
+                doc-builder build ${{ inputs.package }} ../${{ env.doc_folder }}/$lang --build_dir ../build_dir --clean --html ${{ inputs.additional_args }} --language $lang
```
sgugger, line 116: This breaks all the documentations of the libraries, where the source files are in docs/source/, which doesn't end with `en`, so we can't make that change without a broader concertation with the other opensource teams.

lewtun: Indeed, that's what I observed with the way the Spanish translation is currently implemented in the `transformers` docs. 

An alternative would be to provide a mapping between the language and the path where the docs live, e.g. in `transformers` this could be:

```
{"en":"path/to/docs/source", "es":"path/to/docs/source_es"}
```

But that's kind of clunky because we also have args like `path_to_docs` which is used to set `env.doc_folder`. Did you already have something in mind for handling this in `transformers`?

sgugger: We can agree to use that format ultimately (have a folder per language in `docs/source`, I think it makes sense. But it should be opt-in so here `lang` should default to an empty string and when it's an empty string, we should launch the same command as before. This way when a given repo wants to switch to building several languages in the doc, they can do so by reorganizaing the doc folder and providing the argument to the doc building job.

Does that make sense?

lewtun: I like your idea to opt-in! Will adjust the PR accordingly :)

.github/workflows/build_main_documentation.yml
```diff
@@ -95,15 +99,29 @@ jobs:
...
 
       - name: Make documentation
+        shell: bash
         env:
           NODE_OPTIONS: --max-old-space-size=6656
         run: |
           echo "doc_folder has been set to ${{ env.doc_folder }}"
 
-          if [ -z "${{ inputs.notebook_folder }}" ]
+          if [ -z "${{ inputs.notebook_folder }}" ];
```
lewtun, line 108: Now that I've set `shell: bash` we need to explicitly add the comma in the if statement

sgugger, line 108: We're reaching the point where this should be tested on the side and add something to an `EXTRA_ARGS` command, cause all your stuff about language will also need to be in the lese branch of this test ;-)

sgugger (approved):
Better as an opt-in, thanks a lot for adapting your PR!

.github/workflows/build_main_documentation.yml
lewtun: I've now grouped all the base arguments to `doc-builder` in an `args` command here: https://github.com/huggingface/doc-builder/pull/164/commits/dd5d190b2545d76d582a01defcfb63d2036a5faf

I then update `args` if the user provides the `notebook_folder` input and the result is now much simpler :)

The test builds for the [`course`](https://github.com/huggingface/course/runs/5743887075?check_suite_focus=true) and [`datasets` ](https://github.com/huggingface/datasets/runs/5743888058?check_suite_focus=true) PRs still pass, so the last check is to make sure a build of the main docs also works.

lewtun: Here's a test run that we can still build the `datasets` docs with `build_main_documentation.yml` (looks good): https://github.com/huggingface/datasets/runs/5744097672?check_suite_focus=true

.github/workflows/build_main_documentation.yml
```diff
@@ -95,22 +99,34 @@ jobs:
...
+          cd doc-builder
+          args="--build_dir ../build_dir --clean --html ${{ inputs.additional_args }}"
 
-          if [ -z "${{ inputs.notebook_folder }}" ]
+          if [ ! -z "${{ inputs.notebook_folder }}" ];
           then
-            cd doc-builder &&
-            doc-builder build ${{ inputs.package }} ../${{ env.doc_folder }} --build_dir ../build_dir --clean --html ${{ inputs.additional_args }} &&
-            cd ..
+            args="$args --notebook_dir ../notebook_dir"
```
sgugger, line 112: Much cleaner!

sgugger (approved):
Very nice, thanks again!

PR #166, opened 2022-03-29
Fix doc-building on Transformers

sgugger:
#163 introduced a bug that makes the build of Transformers fail. This PR fixes it by adding an extra check to avoid introducing `None` anchors.


LysandreJik (approved):
LGTM!

PR #167, opened 2022-03-30
Add note on multilingual docs to README

lewtun:


sgugger (commented):
Thanks for documenting this, but it's in the wrong place. This should go under a new section detailing how to use the templates for doc building, not inside the manual doc building section.

lewtun:
Thanks for the suggestion! I've migrated the new content to its own section :)

sgugger (approved):
Super nice!

PR #169, opened 2022-03-30
Enable Hub docs directory to be configurable

lewtun:
As discussed in https://github.com/huggingface/moon-landing/pull/2499 this PR enables the docs directory on hf.co to be configurable via a `hub_docs_dir` argument. By default, the docs URL will be 

```
# Compatible with existing URL
https://moon-ci-docs.huggingface.co/docs/{package_name}/pr_{pr_number}
```

but for the course we can configure this to be 

```
# New support for the course
https://moon-ci-docs.huggingface.co/course/{package_name}/pr_{pr_number}
```

A test run with the course CI can be found here: https://github.com/huggingface/course/pull/68

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -20,7 +20,7 @@ on:
       languages:
       # supply space-separated language codes
         type: string
-      hub_docs_url:
+      hub_base_path:
         type: string
 
 jobs:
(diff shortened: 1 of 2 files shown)


sgugger (commented):
Thanks for tackling this! That works, but I was wondering if we shouldn't have an arg for the whole endpoint (maybe some other packages will serve on a different url?). Like for the languages it would have a default to "" and if empty, we would resolve it to the current url.

Btw, the default on the languages can be removed, it's `""` (it was a comment in your PR for that but it got drowned in everything else apparently)

lewtun:
Ah nice idea, that's a good way to future-proof ourselves :) Will adjust this PR accordingly.

> Btw, the default on the languages can be removed, it's `""` (it was a comment in your PR for that but it got drowned in everything else apparently)
Sorry I missed that - I think GitHub was having a bad day during that PR :). Does the same apply to `additional_args` which also has `default: ''`?

.github/workflows/build_pr_documentation.yml
```diff
@@ -17,11 +17,11 @@ on:
       # supply --not_python_module for HF Course
       additional_args:
         type: string
-        default: ''
       languages:
       # supply space-separated language codes
         type: string
-        default: ''
+      hub_base_path:
```
lewtun, line 23: I chose to support just the base URL as I couldn't quite figure out how to propagate the other input parameters at this part of the workflow (the GitHub [docs](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-inputs-and-outputs-with-an-action) are a bit bare)

sgugger:
> Sorry I missed that - I think GitHub was having a bad day during that PR :). Does the same apply to additional_args which also has default: ''?

Yup!

lewtun:
Alright, I've adjusted the PR to now use a `hub_base_path` argument and got rid of the spurious default values for `additional_args` and `languages` :)

sgugger (approved):
Thanks a lot! Can we quickly test it before merging on the course doc?

lewtun:
> Thanks a lot! Can we quickly test it before merging on the course doc?

Here's a test PR for the course which works as we'd like: https://github.com/huggingface/course/pull/68#issuecomment-1083528865

sgugger:
Perfect, good to merge then!

PR #174, opened 2022-03-31
Enable support for notebooks in multiple languages

sgugger:
This PR makes sure that when we set a language argument, we build the notebook in subfolders of notebook_dir (one per language). This will change the current structure (e.g. notebooks will now be pushed to transformers_doc/en instead of just transformers_doc) but makes things support multilingual doc, so I think it's okay.


LysandreJik (approved):
LGTM!

PR #175, opened 2022-03-31
Fix space eaten by links not in anchors mapping

sgugger:
When there is a link to an object that exists but is not in the anchor mapping (as seen a lot in the Spanish doc), a space is eaten. This PR fixes that.


mishig25 (approved):
lgtm! thanks for adding this

Issue #173, opened 2022-03-31
Additional directives that would make development easier


LysandreJik:
The following directives would have been helpful when working on `huggingface_hub`:

- A directive to specify in which version it was added (equivalent of [versionadded](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded))
- A directive to specify in which version it was changed (equivalent of [versionchanged](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged))
- A directive to specify it is deprecated (equivalent of [deprecated](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated))

adrinjalali:
Other directives I find usefule:
- `literalinclude`
- `tabularcolumns`
- `note`
- `warning`

mishig25:
@LysandreJik  Please let me know what kind of syntax you'd like?
How does `<xyz number>` look? (ex: `<versionadded  2.5>`, <deprecated 3.1>) ? 

@adrinjalali 
for `note`, we have similar <Tip> object that looks like this:
<img width="500" alt="Screenshot 2022-04-06 at 14 03 10" src="https://user-images.githubusercontent.com/11827707/161970955-c2ac7f9a-f806-4450-80c3-45a76de0254b.png">

for `warning`, we have similar <Tip warning> object that looks like this:
<img width="500" alt="Screenshot 2022-04-06 at 14 04 22" src="https://user-images.githubusercontent.com/11827707/161971105-2dd4d947-a7f7-4c3f-b2da-3819eeda0745.png">

for `tabularcolumns `, since .mdx files support both markdown & html, one can supply markdown or html table (with its custom styles)

for `literalinclude`, I don't have a good idea on how it should be done (cc: @sgugger )

sgugger:
It doesn't look like we can easily do everything `literalinclude` does in just one command. What was the most useful feature you wanted @adrinjalali ? Highlighting lines? Showing line numbers?

adrinjalali:
`literalinclude` is useful when you have python examples (a .py file) and then you want to reference a section of that in a user guide (an .mdx file). It's only useful once we have the "compile a .py into an .mdx" kinda thing working.

LysandreJik:
This 

> How does <xyz number> look? (ex: <versionadded 2.5>, <deprecated 3.1>) ?

looks good to me!

albertvillanova:
[...]

PR #179, opened 2022-04-08
Literalinclude start-after end-before

mishig25:
I'm in the process of migrating [tokenizers](https://github.com/huggingface/tokenizers) & [inference-api ](https://api-inference.huggingface.co/docs/python/html/index.html) repos to new frontend.

One of the missing capabilities needed is `literalinclude start-after end-before` (example [here](https://github.com/huggingface/tokenizers/blob/main/docs/source/pipeline.rst#id24)), as mentioned in #173

The syntax I'm proposing is html-like tags `<literalinclude>...</literalinclude>`, which will hold python dict str between them specifying options. Wdyt?

Example:
```
<literalinclude>
{'path': './data/convert_literalinclude_dummy.txt', # relative path
'language': 'python', # defaults to '' (empty str)
'start-after': 'START python_import',
'end-before': 'END python_import',
'dedent': 7 # defaults to 0
}
</literalinclude>
```

cc: @adrinjalali @Narsil

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,8 +208,8 @@ There should be json between `<literalinclude>` open & close tags.
 <literalinclude>
 {"path": "./data/convert_literalinclude_dummy.txt", # relative path
 "language": "python", # defaults to " (empty str)
-"start-after": "START python_import",
-"end-before": "END python_import",
+"start-after": "START python_import",  # defaults to start of file
+"end-before": "END python_import",  # defaults to end of file
 "dedent": 7 # defaults to 0
 }
 </literalinclude>
--- a/README.md
+++ b/README.md
@@ -202,6 +202,18 @@ by default) you can put the list of methods to add in a list that contains `all`
     - __call__
 ```
 
+You can create a code-block by referencing a file excerpt with `<literalinclude>` (sphinx-inspired) syntax. 
+```
+<literalinclude>
+{'path': './data/convert_literalinclude_dummy.txt', # relative path
+'language': 'python', # defaults to '' (empty str)
+'start-after': 'START python_import',
+'end-before': 'END python_import',
+'dedent': 7 # defaults to 0
+}
+</literalinclude>
+```
+
 ### Writing source documentation
 
 Arguments of a function/class/method should be defined with the `Args:` (or `Arguments:` or `Parameters:`) prefix, followed by a line return and
--- a/README.md
+++ b/README.md
@@ -203,13 +203,14 @@ by default) you can put the list of methods to add in a list that contains `all`
 ```
 
 You can create a code-block by referencing a file excerpt with `<literalinclude>` (sphinx-inspired) syntax. 
+There should be json between `<literalinclude>` open & close tags.
 ```
 <literalinclude>
-{'path': './data/convert_literalinclude_dummy.txt', # relative path
-'language': 'python', # defaults to '' (empty str)
-'start-after': 'START python_import',
-'end-before': 'END python_import',
-'dedent': 7 # defaults to 0
+{"path": "./data/convert_literalinclude_dummy.txt", # relative path
+"language": "python", # defaults to " (empty str)
+"start-after": "START python_import",
+"end-before": "END python_import",
+"dedent": 7 # defaults to 0
 }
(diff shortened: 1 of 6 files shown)


Narsil:
I am not sure about who's responsible of what.

In HTML terms I think the litteral include could be extracted while the docs are building maybe, so that there's no necessity to depend on other files after building. (Not sure if this is what you propose or not).

`litteral-include` for `tokenizers` and the API serve mostly as a way to ensure the quoted code is actually tested against regularly.
The ability to test examples in `transformers` might be another way forward to avoid `litteral-include`. (But both API and `tokenizers`  also test non Python languages like Node or Curl)

mishig25:
> In HTML terms I think the litteral include could be extracted while the docs are building maybe, so that there's no necessity to depend on other files after building. (Not sure if this is what you propose or not).

let's say you have a file `test.mdx` that has lines:
```
# Some Heading

Some paragraph 

<literalinclude>
{'path': './data/some_source.py', # relative path
'language': 'python', # defaults to '' (empty str)
'start-after': 'START python_import',
'end-before': 'END python_import',
'dedent': 7 # defaults to 0
}
</literalinclude>
```

then, `test.mdx` get transformed into updated `test.mdx` markdown that no longer depends on the `./data/some_source.py` :


# Some Heading

Some paragraph 

````python
some python code excerpt
````

please let me know if this answers your question.


> But both API and tokenizers also test non Python languages like Node or Curl)

indeed, which is why I we may need this < literalinclude >. Otherwise, pydoctests would be suffieicnet

**Comment by Narsil** [comment]
Ok I understood !

Fine by me !

File: src/doc_builder/convert_md_to_mdx.py [hunk 1]
```diff
@@ -101,6 +101,41 @@ def clean_doctest_syntax(text):
...
 
+_re_literalinclude = re.compile(r"([ \t]*)<literalinclude>(((?!<literalinclude>).)*)<\/literalinclude>", re.DOTALL)
+
+
+def convert_literalinclude_helper(match, page_info):
+    """
+    Convert a literalinclude regex match into markdown code blocks by opening a file and
+    copying specificed start-end section into markdown code block.
+    """
+    literalinclude_info = eval(match[2].strip())
```
sgugger, line 112: This will evaluate any code any user has put between litteral include tags on any of their PRs, so not super safe. Could we use something like `json.loads` that is safer?

README.md
```diff
@@ -202,6 +202,18 @@ by default) you can put the list of methods to add in a list that contains `all`
     - __call__
 ```
 
+You can create a code-block by referencing a file excerpt with `<literalinclude>` (sphinx-inspired) syntax. 
+```
+<literalinclude>
+{'path': './data/convert_literalinclude_dummy.txt', # relative path
+'language': 'python', # defaults to '' (empty str)
+'start-after': 'START python_import',
+'end-before': 'END python_import',
```
sgugger, line 211: Should those default to start and end of file, or always have to be provided?

sgugger (commented):
Thanks a lot for working on this! My main concern is with the `eval` you are using, which is very unsafe for jobs that will run in our GH actions (some with push rights on our repos). We should use an alternative :-)

README.md
mishig25: i'm happy to default them to `start and end of file`

src/doc_builder/convert_md_to_mdx.py
mishig25: indeed :) working on it

README.md
mishig25: handled in https://github.com/huggingface/doc-builder/pull/179/commits/8bead14d4b1538ce68fd4a537dfa8ab240f99ede

src/doc_builder/convert_md_to_mdx.py
mishig25: handled in https://github.com/huggingface/doc-builder/pull/179/commits/7a64237749cbbd608a5f0523b8456ead97718429

sgugger (approved):
Nice, thanks for adding my quirks :-)

PR #184, opened 2022-04-12
Raise exception when non-unique headings

mishig25:


src/doc_builder/build_doc.py
```diff
@@ -282,6 +283,9 @@ def generate_frontmatter_in_text(text, file_name=None):
             local = re.sub(r"[^a-z0-9\s]+", "", title.lower())
             local = re.sub(r"\s{2,}", " ", local.strip()).replace(" ", "-")
         text[idx] = f'<h{header_level} id="{local}">{title}</h{header_level}>'
+        if local in locals_encountered:
+            raise ValueError(f"{file_name} contains non-unique (duplicate) headings '{local}'")
```
sgugger, line 287: Maybe not raise just yet but group all errors to raise at the end of the loop? This way we will tell the user once for all duplicates.

sgugger (commented):
Thanks for working on this!

src/doc_builder/build_doc.py
mishig25: great suggestion! handled in https://github.com/huggingface/doc-builder/pull/184/commits/79a8d7ed9d79366060e5a426db11053e53a5a493

sgugger (approved):
Very nice!

PR #188, opened 2022-04-13
Changes needed for tokenizers

mishig25:
2 features are turned off for tokenizers doc builds:
1. Tokenizers python docs will NOT have `source` btn 
   Screenshot 2022-04-13 at 12 37 29
2.`doc-builder preview` cmd hot reloading will NOT track docstring changes in tokenizers python bindings files.

The reason why these 2 features are turned off are: tokenizers python bindings objs:
1.  do not work with `inspect.getsourcelines` & `inspect.getfile`
2. do not have `__module__` attribute 

an error gets thrown:
```
<function descriptor> does not have XYZ
```

taking the easier/simpler route of excluding those 2 features (imo, since they are not essential, especiialy the second one) rather than trying to find a solution [because](https://huggingface.slack.com/archives/C02GLJ5S0E9/p1649327824975459?thread_ts=1649326613.932349&cid=C02GLJ5S0E9)
>  The tokenizers doc is (sadly) not super used, so let's not make too many customizations to doc-builder for it

cc: @Narsil @julien-c

### DIFF
--- a/kit/preprocess.js
+++ b/kit/preprocess.js
@@ -25,13 +25,10 @@ export const docstringPreprocess = {
 			const name = docstringBody.match(REGEX_NAME)[1];
 			const anchor = docstringBody.match(REGEX_ANCHOR)[1];
 			const signature = docstringBody.match(REGEX_SIGNATURE)[1];
-			const source = docstringBody.match(REGEX_SOURCE)[1];
 
 			let svelteComponent = `<Docstring name={${JSON.stringify(
 				unescapeUnderscores(name)
-			)}} anchor={${JSON.stringify(anchor)}} parameters={${signature}} source={${JSON.stringify(
-				source
-			)}} `;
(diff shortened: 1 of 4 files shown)


Narsil:
> taking the easier/simpler route of excluding those 2 features (imo, since they are not essential, especiialy the second one) rather than trying to find a solution [because](https://huggingface.slack.com/archives/C02GLJ5S0E9/p1649327824975459?thread_ts=1649326613.932349&cid=C02GLJ5S0E9)

Very reasonable choice. Also very hard to achieve anyway since the real logic is in Rust code.

sgugger (approved):
Fine with the two choices, it's the right move for now. We can always revisit in the future of course!

Thanks a lot for working on this!

mishig25:
@Narsil @sgugger thanks a lot for the support!

PR #191, opened 2022-04-18
Autodoc `getset_descriptor`

mishig25:
(it should be the last tokenizrs related change 😊)

afaik, [tokenizers @properties](https://github.com/huggingface/tokenizers/blob/88d718207a083795e4b40d80bdc115bd0c065c57/bindings/python/py_src/tokenizers/__init__.pyi#L952-L976) are interpreted as of type `getset_descriptor` (likely because they are bindings) & `getset_descriptor` does not have [fget](https://github.com/huggingface/doc-builder/blob/dd36a89e271bf21be6acd6526a65d3545f318864/src/doc_builder/autodoc.py#L339-L341). Therefore, I'm adding code that makes them posible to use [document_object](https://github.com/huggingface/doc-builder/blob/dd36a89e271bf21be6acd6526a65d3545f318864/src/doc_builder/autodoc.py#L320) with `getset_descriptor` objects.

Note: the other alternative was just to omit `getset_descriptor`  objects from docs. However, imo, they looked too important to omit. Thus, I've decided to handle them.

check examples [here](https://moon-ci-docs.huggingface.co/docs/tokenizers/pr_1/en/api/tokenizer#tokenizers.Tokenizer.decoder)
Screenshot 2022-04-18 at 12 22 55


sgugger (approved):
That's very clear, thanks!
The changes LGTM, thanks! Could you just show what it changes in the doc? (not sure if your screenshot if before or after).

mishig25:
before: [document_object](https://github.com/huggingface/doc-builder/blob/dd36a89e271bf21be6acd6526a65d3545f318864/src/doc_builder/autodoc.py#L320) method errs when object type is `getset_descriptor`

after: [document_object](https://github.com/huggingface/doc-builder/blob/dd36a89e271bf21be6acd6526a65d3545f318864/src/doc_builder/autodoc.py#L320) produces svelte frontend that looks like this: starts with `@property`
Screenshot 2022-04-18 at 12 22 55

please let me know if the description is clear 😊

sgugger:
Just wondering where the `@proprerty` comes from (especially the `@`). Is this something we use elsewhere already? I think just `property` is enough no?

mishig25:
For class docstrings we have, `class`
Screenshot 2022-04-18 at 14 34 43
For function/method docstrings we have, `f`
Screenshot 2022-04-18 at 14 35 02
and I wanted to have something equavlanet for `properties`L with `@property`:
in https://github.com/huggingface/doc-builder/pull/191/commits/6b41981a710a2c1fabd2515a1f9eeaebd63ab0c7, I've changed from `@property` to just `property` as suggested

sgugger:
We could have a stylized icon for them too!

mishig25:
@gary149 any icn suggestions for [python property](https://www.programiz.com/python-programming/property) ?

mishig25:
@sgugger after getting input from @gary149 this is how it looks [here](https://moon-ci-docs.huggingface.co/docs/tokenizers/pr_1/en/api/tokenizer#tokenizers.Tokenizer.truncation)

`@` icon and green (different) color so that we can quickly differentiate between function vs property

Screenshot 2022-04-19 at 12 24 14

I'm gonna merge it if you dont have any objection 😊

sgugger:
Super nice, thanks for iterating on the design!

PR #192, opened 2022-04-19
Make `pip install .[dev]` path configurable

mishig25:


mishig25:
Supercedede by https://github.com/huggingface/doc-builder/pull/194

PR #194, opened 2022-04-21
GH workflow changes for tokenizers

mishig25:
Needed for https://github.com/huggingface/tokenizers/pull/980

See [here](https://github.com/huggingface/tokenizers/blob/95b5d066d5b90582ceccb76b84b7d41351532e82/.github/workflows/build_pr_documentation.yml#L18-L19)

Adding 2 new [inputs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow) to doc build gh workflows:
1. `package_path`: because tokenizers python package is not in its git root folder, but in `tokenizers/bindings/python`
2. `install_rust`: installing rust is needed to install tokenizers from source

For install `install_rust`, I wanted to re-use [additional_args](https://github.com/huggingface/doc-builder/blob/main/.github/workflows/build_pr_documentation.yml#L18) but `additional_args` gets passed to `doc-builder`, which errors:
```bash
Make documentation step
args="--build_dir ../build_dir --clean --version pr_980 --html --rust-toolchain"
Doc Builder CLI tool: error: unrecognized arguments: --rust-toolchain
```

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -19,6 +19,10 @@ on:
       languages:
       # supply space-separated language codes
         type: string
+      package_path:
+        type: string
+      install_rust:
+        type: boolean
     secrets:
       token:
         required: true
@@ -67,6 +71,12 @@ jobs:
             echo "doc_folder=${{ inputs.path_to_docs }}" >> $GITHUB_ENV
           fi
 
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        if: inputs.install_rust
+        with:
+          toolchain: stable
+
       - name: Setup environment
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


lewtun (approved):
LGTM!

.github/workflows/build_main_documentation.yml
```diff
@@ -76,7 +86,11 @@ jobs:
           pip install .
           cd ..
 
-          if [[ "${{ inputs.additional_args }}" != *"--not_python_module"* ]];
+          if [ "${{ inputs.package_path }}" ]
```
sgugger, line 89: Don't we need a `-n` here?
```suggestion
          if [[ -n "${{ inputs.package_path }}" ]]
```

.github/workflows/build_pr_documentation.yml
```diff
@@ -76,7 +86,11 @@ jobs:
           pip install .
           cd ..
 
-          if [[ "${{ inputs.additional_args }}" != *"--not_python_module"* ]];
+          if [ "${{ inputs.package_path }}" ]
```
sgugger, line 89: Same here
```suggestion
          if [[ -n "${{ inputs.package_path }}" ]]
```

sgugger (approved):
Thanks for adding this! Not too sure about the bash syntax myself, but Google says nothing about just putting the variable without using `! -z` or `-n`

LysandreJik (approved):
If you've verified it works, LGTM!

mishig25:
when I tried locally, both `if [ "${{ inputs.package_path }}" ]` & `if [[ -n "${{ inputs.package_path }}" ]]` options worked. However, using `if [[ -n "${{ inputs.package_path }}" ]]` since it looks more "standard"

PR #195, opened 2022-04-22
Fix fw content bug

mishig25:
There was a bug where empty frameworkcontent was being shown (see attached img). The expected behaviours is that: empty framework content should not be displayed

image (5)

### DIFF
--- a/kit/preprocess.js
+++ b/kit/preprocess.js
@@ -133,13 +133,14 @@ export const frameworkcontentPreprocess = {
 		const REGEX_PYTORCH = /<pt>(((?!<pt>).)*)<\/pt>/ms;
 		const REGEX_TENSORFLOW = /<tf>(((?!<tf>).)*)<\/tf>/ms;
 		const REGEX_JAX = /<jax>(((?!<jax>).)*)<\/jax>/ms;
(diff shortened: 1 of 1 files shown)


kit/preprocess.js
```diff
@@ -133,13 +133,14 @@ export const frameworkcontentPreprocess = {
...
 		const REGEX_JAX = /<jax>(((?!<jax>).)*)<\/jax>/ms;
-		const FRAMEWORKS = [
-			{ framework: "pytorch", REGEX_FW: REGEX_PYTORCH, isExist: false },
-			{ framework: "tensorflow", REGEX_FW: REGEX_TENSORFLOW, isExist: false },
-			{ framework: "jax", REGEX_FW: REGEX_JAX, isExist: false }
-		];
-
+		
 		content = await replaceAsync(content, REGEX_FRAMEWORKCONTENT, async (_, fwcontentBody) => {
+			const FRAMEWORKS = [
```
mishig25, line 138: the bug was being caused where `FRAMEWORKS` const was a "global" state rather than "local" (i.e. inside the closure)

gary149 (approved):

PR #201, opened 2022-04-26
Fix source link for objects defined in an __init__

sgugger:
This takes care of the problem highlighted in [this issue](https://github.com/huggingface/transformers/issues/16904). Objects defined in an `__init__.py` have a wrong source link returned.


mishig25 (approved):
Lgtm!

PR #200, opened 2022-04-26
Fix dstasets version `v` tag issue

mishig25:
Closes #189

Creating a new variable `version_tag` (I think it is the simplest solution)

PR #202, opened 2022-04-27
Fix version_tag

mishig25:
There was a bug in `version_tag` was decalred

followup to https://github.com/huggingface/doc-builder/pull/200


src/doc_builder/commands/build.py
```diff
@@ -82,7 +82,8 @@ def build_command(args):
     if args.version is None and "dev" not in version:
         doc_config = get_doc_config()
         version_prefix = getattr(doc_config, "version_prefix", "v")
-        version_tag = f"{version_prefix}{version}"
```
sgugger, line 85: I can't comment above but at this stage `version` won't have `dev` in it anymore if it's the main version. The test a line 82 should become `version != default_version`

PR #203, opened 2022-04-27
Hotfix version issue

mishig25:
Hotfix to [this problem](https://huggingface.slack.com/archives/C027NLU6CE9/p1651056613127479)

Thanks a lot[ for the suggestion](https://github.com/huggingface/doc-builder/pull/202#discussion_r859696270), @sgugger !

(merging them so that I can fix the broken links in the prod 😊)

PR #204, opened 2022-04-27
Rm unneded if condition for `version_tag`

mishig25:
As suggested [here](https://github.com/huggingface/doc-builder/pull/203#discussion_r859752858)

apologies for causing bunch of issues because of #200 😊

Tested both on transformers & datasets, the source urls are working correctly:
https://huggingface.co/docs/transformers/main/en/model_doc/bert#transformers.BertConfig

https://huggingface.co/docs/datasets/master/en/package_reference/main_classes#datasets.DatasetInfo.from_directory


sgugger (approved):
Thanks for fixing!

PR #214, opened 2022-05-18
Commit files to `doc-build(-dev)?` using Github GraphQL

mishig25:
Problem: read more [here](https://huggingface.slack.com/archives/C02GLJ5S0E9/p1648134906365829)
> Cloning it in a GitHub Action workflow takes about 4 minutes, even with a --depth 1

Solution: as suggested by ~@LysandreJik~ @julien-c (edit by @LysandreJik), use Github GraphQL [createcommitonbranch mutation](https://github.blog/changelog/2021-09-13-a-simpler-api-for-authoring-commits/). And this PR does that

I've tested it on accelerate [here](https://github.com/huggingface/accelerate/pull/372/files) and it is [working here](https://github.com/huggingface/accelerate/runs/6498426135?check_suite_focus=true)
btw, can confirm that pushing with GraphQL is way fater since we do not need to clone doc-build-dev (which is roughly 30GB in size atm)

todo (note for myself):
- [ ] revert to `main` branch use in `ref` for the doc-builder checkout

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -71,6 +71,13 @@ jobs:
             echo "hub_docs_url=${{ inputs.hub_base_path }}/${{ inputs.package }}/pr_${{ inputs.pr_number }}" >> $GITHUB_ENV
           fi
 
+          if [ -z "${{ inputs.package_name }}" ];
+          then
+            echo "package_name=${{ inputs.package }}" >> $GITHUB_ENV
+          else
+            echo "package_name=${{ inputs.package_name }}" >> $GITHUB_ENV
+          fi
+
       - name: Install Rust
         uses: actions-rs/toolchain@v1
         if: inputs.install_rust
@@ -133,12 +140,7 @@ jobs:
           cd doc-build-dev
           git pull
           rm -rf ${{ inputs.package }}/pr_${{ inputs.pr_number }}
-          if [ -z "${{ inputs.package_name }}" ];
-          then
-            mv ../build_dir/${{ inputs.package }}/pr_${{ inputs.pr_number }} ${{ inputs.package }}/pr_${{ inputs.pr_number }}
-          else
-            mv ../build_dir/${{ inputs.package }}/pr_${{ inputs.pr_number }} ${{ inputs.package_name }}/pr_${{ inputs.pr_number }}
-          fi
+          mv ../build_dir/${{ inputs.package }}/pr_${{ inputs.pr_number }} ${{ env.package_name }}/pr_${{ inputs.pr_number }}
           git status
 
           if [[ `git status --porcelain` ]]; then
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -29,9 +29,6 @@ on:
         type: string
       install_rust:
         type: boolean
-    secrets:
-      token:
-        required: true
 
 jobs:
   build_pr_documentation:
@@ -143,7 +140,7 @@ jobs:
           if [ "${{ inputs.package }}" != "${{ env.package_name }}" ]; then
             mv ${{ inputs.package }} ${{ env.package_name }}
           fi
-          doc-builder commit ${{ env.package_name }} --doc_build_repo_id "huggingface/doc-build-dev" --token ${{ secrets.token }} --commit_msg "Updated with commit ${{ inputs.commit_sha }} See: https://github.com/huggingface/${{ inputs.package }}/commit/${{ inputs.commit_sha }}"
+          doc-builder commit ${{ env.package_name }} --doc_build_repo_id "huggingface/doc-build-dev" --token ${{ env.write_token }} --commit_msg "Updated with commit ${{ inputs.commit_sha }} See: https://github.com/huggingface/${{ inputs.package }}/commit/${{ inputs.commit_sha }}"
 
       - name: Find doc comment
         uses: peter-evans/find-comment@v1
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -44,6 +44,7 @@ jobs:
         with:
           repository: 'huggingface/doc-builder'
           path: doc-builder
+          ref: use_gh_graphql
 
       - uses: actions/checkout@v2
         with:
@@ -92,7 +93,7 @@ jobs:
         run: |
           pip uninstall -y doc-builder
           cd doc-builder
-          git pull origin main
+          git pull origin use_gh_graphql
           pip install .
           cd ..
 
@@ -136,11 +137,13 @@ jobs:
           cd ..
 
       - name: Push to repositories
-        run: |
-          mv ./build_dir/${{ inputs.package }}/pr_${{ inputs.pr_number }} ${{ env.package_name }}/pr_${{ inputs.pr_number }}
-          doc-builder commit ${{ env.package_name }} --doc_build_repo_id "huggingface/doc-build-dev" --token ${{ secrets.token }} --commit_msg "Updated with commit ${{ inputs.commit_sha }} See: https://github.com/huggingface/${{ inputs.package }}/commit/${{ inputs.commit_sha }}"
-
         shell: bash
+        run: |
+          cd build_dir
+          if [ "${{ inputs.package }}" != "${{ env.package_name }}" ]; then
+            mv ${{ inputs.package }} ${{ env.package_name }}
+          fi
+          doc-builder commit ${{ env.package_name }} --doc_build_repo_id "huggingface/doc-build-dev" --token ${{ secrets.token }} --commit_msg "Test commit GraphQL"
 
       - name: Find doc comment
         uses: peter-evans/find-comment@v1
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -143,7 +143,7 @@ jobs:
           if [ "${{ inputs.package }}" != "${{ env.package_name }}" ]; then
             mv ${{ inputs.package }} ${{ env.package_name }}
           fi
-          doc-builder commit ${{ env.package_name }} --doc_build_repo_id "huggingface/doc-build-dev" --token ${{ secrets.token }} --commit_msg "Test commit GraphQL"
+          doc-builder commit ${{ env.package_name }} --doc_build_repo_id "huggingface/doc-build-dev" --token ${{ secrets.token }} --commit_msg "Updated with commit ${{ inputs.commit_sha }} See: https://github.com/huggingface/${{ inputs.package }}/commit/${{ inputs.commit_sha }}"
 
       - name: Find doc comment
         uses: peter-evans/find-comment@v1
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -105,11 +105,6 @@ jobs:
           fi
           cd ..
 
-      - name: Setup git
-        run: |
-          git config --global user.name "Hugging Face Doc Builder"
-          git config --global user.email docs@huggingface.co
-
       - name: Make documentation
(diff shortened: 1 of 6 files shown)


src/doc_builder/commands/commit.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    res = requests.get(
+        f"https://api.github.com/repos/{repo_id}/git/refs/heads/${branch}",
+        headers={"Authorization": f"bearer {token}"},
+    )
+    if res.status_code == 200:
+        head_oid = _re_head_oid.search(res.text)[1]
+        return head_oid
+    else:
+        raise Exception("get_head_oid failed: {{status: {res.status}, msg:{res.text}}}")
```
julien-c, line 42: why not parse the JSON btw?

src/doc_builder/commands/commit.py
```diff
@@ -0,0 +1,145 @@
...
+        expectedHeadOid: "{head_oid}"
+      }}
+    ) {{
+      clientMutationId
+    }}
+  }}
+  """
+    )
+    # Execute the query on the transport
+    result = client.execute(query)
```
julien-c, line 99: might be better to use graphql "variables" and not create the string dynamically

julien-c: (sorry i know this is not ready for review but I was curious to take a look 😅)

src/doc_builder/commands/commit.py
mishig25: handled in https://github.com/huggingface/doc-builder/pull/214/commits/6ce101dfed4531ac19b405cdf86a3de5df44008d

I tend to over-use regex 😊

src/doc_builder/commands/commit.py
mishig25: handled in https://github.com/huggingface/doc-builder/pull/214/commits/a3ffc278896381771a5a923c7db362de29dd24be
using "variables" actually make the implemntation cleaner, great suggestion!

.github/workflows/build_pr_documentation.yml
```diff
@@ -29,6 +29,9 @@ on:
         type: string
       install_rust:
         type: boolean
+    secrets:
+      token:
+        required: true
```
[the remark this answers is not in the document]
mishig25: Oh I see, thanks a lot for catching this bug.
handled in https://github.com/huggingface/doc-builder/pull/214/commits/d41d23ce82bed0d72b0d3239012535b22ada0372 and tested in accelerate https://github.com/huggingface/accelerate/runs/6509119836?check_suite_focus=true

src/doc_builder/commands/commit.py
```diff
@@ -0,0 +1,145 @@
...
+        fileChanges: {
+          additions: $additions
+        },
+        expectedHeadOid: $head_oid
+      }
+    ) {
+      clientMutationId
+    }
+  }
+  """
```
sgugger, line 92: Might be worth putting into a constant?

src/doc_builder/commands/commit.py
```diff
@@ -0,0 +1,145 @@
...
+    result = client.execute(query, variable_values=params)
+    return result
+
+
+def commit_command(args):
+    """
+    Commit file additions using Github GraphQL rather than `git`.
+    Usage: doc-builder commit $args
+    """
+    number_of_retries = 5
```
sgugger, line 105: Should be one of the accepted args maybe?

sgugger (approved):
I don't know enough about GitHub GraphQL to fully review that part, but it looks to be working and be waaaay faster. Left a few comments on the general API.

Re-naming: the command actually commits and pushes, so `doc-builder push` or `doc-builder commit-and-push` might be more accurate than `doc-builder commit`?

src/doc_builder/commands/commit.py
mishig25: handled in https://github.com/huggingface/doc-builder/pull/214/commits/2d317cddc6babeef7f5a1492aa00ee77d3673157

mishig25:
as suggested by @sgugger, renamed command `doc-builder commit` -> `doc-builder push`

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,149 @@
...
+        raise ValueError(f"CLI arg `n_retries` MUST be positive & non-zero; supplied value was {args.n_retries}")
+
+    MAX_N_RETRIES = args.n_retries + 1
+    number_of_retries = args.n_retries
+    additions_str = create_additions(args.path_to_built_docs)
+    while number_of_retries:
+        try:
+            head_oid = get_head_oid(args.doc_build_repo_id, args.token)
+            commit_additions(additions_str, args.doc_build_repo_id, head_oid, args.token, args.commit_msg)
+            break
```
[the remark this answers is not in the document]
julien-c: (maybe log/print the time it takes for both `create_additions` and `commit_additions`)

julien-c (commented):
Didn't actually test it, but the code is really really neat

GG @mishig25 !

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    if res.status_code == 200:
+        res_json = res.json()
+        head_oid = res_json["object"]["sha"]
+        return head_oid
+    else:
+        raise Exception(f"get_head_oid failed: {res.message}")
+
+
+def create_additions(path_to_built_docs):
```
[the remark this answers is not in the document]
sgugger: At least in the docstring yes, if not the signature :-)

LysandreJik (approved):
Very clean!

mishig25:
Thank you all for the kind words. Obviously, I am learning from the best team, and excited to keep learning and keep the momentum going!

src/doc_builder/commands/commit.py
mishig25: definitely, @Pierrci's fault 😊

Pierrci: mmmm I'm not that sure haha

mishig25:
When I tested on transformers, I needed this change:
https://github.com/huggingface/doc-builder/blob/59e8f863bc878ed2b62403500c6bcd5f26003710/src/doc_builder/commands/push.py#L70-L75

[Here, you can see output](https://github.com/huggingface/transformers/runs/6555120464?check_suite_focus=true#step:11:19) of a successful run (23 seconds vs ):
```bash
create_additions took 0.3139 seconds or 0.01 mins
Committed additions chunk: 1/4
Committed additions chunk: 2/4
Committed additions chunk: 3/4
Committed additions chunk: 4/4
commit_additions took 21.2038 seconds or 0.35 mins
```

please let me know wdy about this change @LysandreJik @sgugger @julien-c

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,208 @@
...
+
+    return additions
+
+
+MAX_CHUNK_LEN = 3e7  # 30 Megabytes
+
+
+def create_additions_chunks(additions: List[FileAddition]) -> List[List[FileAddition]]:
+    """
+    Github GraphQL `createCommitOnBranch` mutation fails when a payload is bigger than 50 MB.
```
mishig25, line 72: I've found this 50MB limit by experimentally.
I couldn't find anything in the docs https://docs.github.com/en/graphql/overview/resource-limitations

When I created junk payloads for testing that exceeded 50MB, I got error:
```json
{'message': 'Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `C0A9:0BB2:1828F7:1A00FE:6287547B` when reporting this issue.'}
```

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,208 @@
...
+    Therefore, in those cases (transformers doc ~ 100 MB), we need to commit using
+    multiple smaller `createCommitOnBranch` mutation calls.
+    """
+    additions_chunks = []
+    current_chunk = []
+    current_len = 0
+
+    while additions:
+        addition = additions.pop()
```
sgugger, line 81: It doesn't look like `additions` is being modified in this loop, so let's just do a for loop:
```suggestion
    for addition in additions:
```

sgugger (commented):
New chunking system looks good to me!

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,208 @@
...
+    transport = AIOHTTPTransport(
+        url="https://api.github.com/graphql", headers={"Authorization": f"bearer {args.token}"}
+    )
+    gql_client = Client(transport=transport, fetch_schema_from_transport=True, execute_timeout=None)
+
+    # commit file additions
+    time_start = time()
+    additions = create_additions(args.library_name)
+    time_end = time()
+    print(f"create_additions took {time_end-time_start:.4f} seconds or {(time_end-time_start)/60.0:.2f} mins")
```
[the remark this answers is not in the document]
mishig25: I tried to address Julien's comment https://github.com/huggingface/doc-builder/pull/214#discussion_r877376821

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,208 @@
...
+                print(f"Committed additions chunk: {i+1}/{len(additions_chunks)}")
+            break
+        except Exception as e:
+            number_of_retries -= 1
+            print(f"createCommitOnBranch error occurred: {e}")
+            if number_of_retries:
+                print(f"Failed on try #{max_n_retries-number_of_retries}, pushing again")
+
+    time_end = time()
+    print(f"commit_additions took {time_end-time_start:.4f} seconds or {(time_end-time_start)/60.0:.2f} mins")
```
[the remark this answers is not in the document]
mishig25: Same as here I tried to address Julien's comment https://github.com/huggingface/doc-builder/pull/214#discussion_r877376821

src/doc_builder/commands/push.py
```diff
@@ -0,0 +1,208 @@
...
+        raise Exception(f"get_head_oid failed: {res.message}")
+
+    res_json = res.json()
+    head_oid = res_json["object"]["sha"]
+    return head_oid
+
+
+class FileAddition(TypedDict):
+    path: str
+    contents: int
```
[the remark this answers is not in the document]
mishig25: `class FileAddition(TypedDict)` is purely for typing reasons using [TypedDict](https://peps.python.org/pep-0589/)
if I use dataclass, then, I have to change line below to use the dataclass as in `additions.append(new FileAddition(...))`, which would break the graphql https://github.com/huggingface/doc-builder/blob/59e8f863bc878ed2b62403500c6bcd5f26003710/src/doc_builder/commands/push.py#L62

src/doc_builder/commands/push.py
mishig25: great suggestion!

src/doc_builder/commands/push.py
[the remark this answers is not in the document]
sgugger: Maybe use the `logging` module and set those at the DEBUG level then, instead of printing all the time.

src/doc_builder/commands/push.py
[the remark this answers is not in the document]
sgugger: Oh I think you were actually using it.
This is Python and we don't use type checker on any of our repos. I don't think introducing this helps in any way, so I'd remove it as well as the type annotations below (`Dict` is enough honestly)

Also this is only available in Python 3.8+ unless some extra package is installed, which the setup of `doc-builder` doesn't have.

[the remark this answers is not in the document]
mishig25: handled in https://github.com/huggingface/doc-builder/pull/214/commits/5257820158750fe41371a025d5d16501c90ea2f7

src/doc_builder/commands/push.py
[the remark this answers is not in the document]
mishig25: handled in https://github.com/huggingface/doc-builder/pull/214/commits/a4cf6fa61ec1157e9bcd3bf0faaf1a21243f20e6

PR #262, opened 2022-07-12
Don't add () for properties

sgugger:
When resolving links, the doc-builder adds parenthesis for every object that is not a class. But properties shouldn't get a pair. This PR fixes that.


muellerzr (approved):
Thanks!

PR #263, opened 2022-07-12
Add notebook_to_mdx command

sgugger:
This PR adds a command to convert a notebook to an mdx file, making it easier to draft a new tutorial as a notebook and then converting it.


muellerzr (approved):
Thanks for this!

PR #264, opened 2022-07-13
Document how attributes should be included

muellerzr:
This PR documents how to document an enum-like class and not have the attributes render as `Parameters:` in the documentation.


sgugger (approved):
Thanks for documenting this!

PR #273, opened 2022-08-09
Fix bug in styler when line after Parameter block is empty

sgugger:
There is currently a bug in the doc styler when there is an empty line after the parameter introduction:
```py
"""
Function description

Params:

    x (`int`): This is x.
    y (`float`): this is y.
"""
```
This will reformat all arguments on the same line after, an issue we have seen recently in Transformers with a generate doc PR. It will give:
```py
"""
Function description

Params:

    x (`int`): This is x. y (`float`): this is y.
"""
```

This PR fixes the bug and removes the extra empty lines so reformats the above as:
```py
"""
Function description

Params:
    x (`int`): This is x.
    y (`float`): this is y.
"""
```

Note: after being merged, we will need a release and a quick update to Transformers as this changes 16 files currently.

### DIFF
--- a/src/doc_builder/style_doc.py
+++ b/src/doc_builder/style_doc.py
@@ -260,7 +260,9 @@ def style_docstring(docstring, max_len):
     ):
         param_indent = find_indent(lines[idx])
 
-    for idx, line in enumerate(lines):
+    idx = 0
+    while idx < len(lines):
+        line = lines[idx]
         # Doing all re searches once for the one we need to repeat.
         list_search = _re_list.search(line)
         code_search = _re_code.search(line)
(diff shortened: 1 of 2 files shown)


src/doc_builder/style_doc.py
```diff
@@ -316,7 +318,13 @@ def style_docstring(docstring, max_len):
...
         elif _re_args.search(line):
             new_lines.append(line)
-            param_indent = find_indent(lines[idx + 1])
+            idx += 1
+            while idx < len(lines) and is_empty_line(lines[idx]):
+                idx += 1
+            if idx < len(lines):
+                param_indent = find_indent(lines[idx])
+                # We still need to treat that line
+                idx -= 1
```
LysandreJik, line 327: Smart!

LysandreJik (approved):
LGTM!

sgugger:
Merging then, will do a release soon-ish.

PR #278, opened 2022-08-23
Add version suffix option

fxmarty:
Add an option to set the suffix (e.g. by default `src/`) after the version tag.

This will help fixing the broken documentation links in Optimum documentation as the structure is not `src/optimum/` but directly `optimum/` (see e.g. the broken links in https://huggingface.co/docs/optimum/main/en/onnxruntime/quantization#optimum.onnxruntime.ORTQuantizer ). Passing `doc-builder build --version_tag_suffix ""` will fix the issue.

An other approach could be to use an environment variable, let me know if you prefer it this way.

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -325,13 +325,13 @@ def is_getset_descriptor(obj):
     return "getset_descriptor" in obj_repr
 
 
-def get_source_link(obj, page_info):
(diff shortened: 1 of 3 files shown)


mishig25 (approved):
lgtm!

mishig25:
@fxmarty @lewtun please feel free to update the optimum doc-build scripts with `--version_tag_suffix`

PR #295, opened 2022-09-14
Support `build` on Windows

mariosasko:
Use POSIX paths (where needed) to avoid the `build` errors on Windows due to `\` being the path separator. `\` can cause the invalid escape sequence error in the downstream regex patterns (another option would be to use `re.escape` to fix this) and make the path comparison fail in the toc integrity check.


sgugger (approved):
Thanks for fixing this! I'd be surprised if it's the only fix necessary for Windows support though ;-)

mariosasko:
Yes, I should be more precise - this PR contains the fixes needed to build the `datasets` docs on Windows successfully. A better test would be building the docs for `transformers`, but this still fails (mid-way) due to Jax/Flax not being supported, which makes it impossible to resolve the references to the Flax models.

mishig25:
cc: @dylanebert

PR #296, opened 2022-09-14
Optimize example code block regex

mariosasko:
Fix the example code block regex to avoid the so-called "catastrophic backtracking" scenario, which causes CI to hang. For the reproduction, check [this](https://github.com/huggingface/datasets/blob/18e7454d53633577250452616e6991af064ac050/src/datasets/arrow_dataset.py#L959-L963) commit from https://github.com/huggingface/datasets/pull/4957 (notice the missing closing `` ``` ``).


mishig25 (approved):
lgtm! 
tested it & works great!

PR #302, opened 2022-09-26
Add repo_owner input to build workflows

### LINKED ISSUE #300: Add ability to build from repos under different org/user
Right now I'm trying to set up hugging face docs for `timm`, which is under `rwightman`'s username. Would it make sense to add an input in `build_main_documentation.yml` to optionally override the repo owner of the repo you're trying to build documentation for?

Looks like we could do that, then just replace `huggingface` [here](https://github.com/huggingface/doc-builder/blob/da021d60070ff3bc5a33be7fc457656d242ba6cb/.github/workflows/build_main_documentation.yml#L49) with a new input, `${{ inputs.repository_owner }}`, that defaults to `huggingface`.

nateraw:
Resolves #300. Any thoughts on `"required: true"` vs `"required: false"` would be nice here :).

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -28,6 +28,11 @@ on:
         type: boolean
       install_libgl1:
         type: boolean
+      repo_owner:
+        type: string
+        required: true
+        default: 'huggingface'
+        description: "Owner of the repo to build documentation for. Defaults to 'huggingface'."
     secrets:
       token:
         required: true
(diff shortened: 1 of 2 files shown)


mishig25 (approved):
lgtm!

mishig25:
@nateraw thanks a lot for this change!

after some testing, I had to turn remove `required: True` so that it `repo_owner` falssback to `huggingface` default value when non-provided. 
Successfully tested a doc-build [here](https://github.com/huggingface/accelerate/actions/runs/3130077651/jobs/5087819416) & merged [here](https://github.com/huggingface/doc-builder/commit/779ff378d0f6f1d6862c5fa106f338930f970213)

nateraw:
Thank you, sir!!

PR #311, opened 2022-10-07
Resolve params in text

mariosasko:
Resolve function/method parameter links in MD/MDX files.

This is the syntax for the parameter links:
```
[`package.module.function#parameter`]
``` 
And this is what's left after the resolution step:
```
[parameter](url_with_function#package.module.function.parameter)
```

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -542,14 +542,19 @@ def resolve_links_in_text(text, package, mapping, page_info):
             link = get_external_object_link(object_or_param_name, page_info)
             return f"{link}{last_char}"
         object_name, param_name = None, None
+        # If `#` is in the name, assume it's a link to the function/method parameter.
+        if "#" in object_or_param_name:
+            object_name_for_param = object_or_param_name.split("#", 1)[0]
+            # Strip preceding `~` if it's there.
+            object_name_for_param = (
+                object_name_for_param[1:] if object_name_for_param.startswith("~") else object_name_for_param
+            )
+            obj = find_object_in_package(object_name_for_param, package)
+            param_name = object_or_param_name.split("#", 1)[-1]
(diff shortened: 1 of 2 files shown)


src/doc_builder/autodoc.py
```diff
@@ -533,36 +533,47 @@ def resolve_links_in_text(text, package, mapping, page_info):
...
-        if object_name.startswith("~"):
-            obj = find_object_in_package(object_name[1:], package)
-            object_name = object_name.split(".")[-1]
+        if object_or_param_name.startswith("~"):
+            obj = find_object_in_package(object_or_param_name[1:], package)
+            object_name = object_or_param_name.split(".")[-1]
+        # If "#" is in the name, assume it's a link to the function/method parameter.
+        elif "#" in object_or_param_name:
+            obj = find_object_in_package(object_or_param_name.split("#", 1)[0], package)
+            param_name = object_or_param_name.split("#", 1)[-1]
```
sgugger, line 552: This should probably be done before the test at line 546, as it should be possible to do
```
[`~package.module.object#param`]
```

tests/test_autodoc.py
```diff
@@ -371,6 +371,18 @@ def test_resolve_links_in_text(self):
...
+                "Link to [`transformers.BertModel.forward#input_ids`].",
+                transformers,
+                small_mapping,
+                page_info,
+            ),
+            (
+                "Link to [input_ids](/docs/transformers/main/en/model_doc/bert.html#transformers.BertModel.forward.input_ids)."
+            ),
+        )
+
```
sgugger, line 385: Would add the same test here with ``[`~transformers.BertModel.forward#input_ids`]`` to make sure it works too.

sgugger (commented):
Thanks a lot for working on this! I think the case combining a ~ and a # should be supported.

sgugger (approved):
Thanks a lot!

PR #324, opened 2022-11-15
Handle the case where there is no empty line before the argument block

gante:
As the title says, handles the case where there is no empty line before the argument block, which was causing weird patterns to pop up in other libraries (like `transformers`).

For instance, 
```
Some description
Args:
    - foo
```

was being converted to

```
Args:
Some description
    - foo
```

This was because the paragraph was not seen as completed (no blank line), and the `Args:` line was appended as soon as it was seen. After this PR, the example above is converted to 

```
Some description

Args:
    - foo
```

i.e. an empty line before `Args:` is added when it is missing.

### DIFF
--- a/src/doc_builder/style_doc.py
+++ b/src/doc_builder/style_doc.py
@@ -263,9 +263,11 @@ def style_docstring(docstring, max_len):
     idx = 0
     while idx < len(lines):
         line = lines[idx]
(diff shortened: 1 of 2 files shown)


gante:
(@sgugger not sure who else to tag in the PR review :) )

sgugger (approved):
Great! Could you also add a test for it?

PR #334, opened 2022-12-05
Add repo owner to doc source links

nateraw:
While adding some reference docs to `timm`'s Docs, I noticed the "Source" links were incorrect in the API documentation - they always pointed to `github.com/huggingface/timm`. We want them to point to `github.com/rwightman/pytorch-image-models`. 

This PR aims to make that happen. 

As of the time of creating this PR, there is an ambiguity between `package` and `package_name`, which still makes links generate incorrectly with `package_name=timm`, as the github repo name should still be `pytorch-image-models`. Will sort out a fix for that...

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -50,7 +50,6 @@ jobs:
         with:
           repository: 'huggingface/doc-builder'
           path: doc-builder
-          ref: repo-owner-bugfix
 
       - uses: actions/checkout@v2
         with:
@@ -113,7 +112,6 @@ jobs:
         run: |
           pip uninstall -y doc-builder
           cd doc-builder
-          git pull origin repo-owner-bugfix
           pip install .
           cd ..
 
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -139,7 +139,7 @@ jobs:
         run: |
           echo "doc_folder has been set to ${{ env.doc_folder }}"
           cd doc-builder
-          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }} --repo_owner ${{ inputs.repo_owner }} --repo_name ${{ inputs.package }} --version_tag_suffix ${{ inputs.version_tag_suffix }}"
+          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }} --repo_owner ${{ inputs.repo_owner }} --repo_name ${{ inputs.package }} --version_tag_suffix '${{ inputs.version_tag_suffix }}'"
 
           if [ -z "${{ inputs.languages }}" ];
           then
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -134,7 +134,7 @@ jobs:
         run: |
           echo "doc_folder has been set to ${{ env.doc_folder }}"
           cd doc-builder
-          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }}"
+          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }} --repo_owner ${{ inputs.repo_owner }}"
 
           if [ -z "${{ inputs.languages }}" ];
           then
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -139,7 +139,7 @@ jobs:
         run: |
           echo "doc_folder has been set to ${{ env.doc_folder }}"
           cd doc-builder
-          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }} --repo_owner ${{ inputs.repo_owner }} --repo_name ${{ inputs.package }} --version_tag_suffix '${{ inputs.version_tag_suffix }}'"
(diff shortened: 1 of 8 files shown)


nateraw:
@mishig25 do you think I should go with:

1. the route I'm doing here, where I'm parametrizing the repo name/repo owner (probably the right way to do it but I'll have to rewrite some tests and I'm afraid something in another repo will break)
2. Just adding explicit override when package name is `pytorch-image-models` to update github links/package name

nateraw:
@sgugger mind having a look at this and weighing in on the above?

.github/workflows/build_pr_documentation.yml
```diff
@@ -134,7 +135,7 @@ jobs:
         run: |
           echo "doc_folder has been set to ${{ env.doc_folder }}"
           cd doc-builder
-          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }}"
+          args="--build_dir ../build_dir --clean --version pr_${{ inputs.pr_number }} --html ${{ inputs.additional_args }} --repo_owner ${{ inputs.repo_owner }} --repo_name ${{ inputs.package }}"
```
sgugger, line 138: Note that you will also need to change the command in the `build_documentation` job.

src/doc_builder/autodoc.py
```diff
@@ -329,9 +329,10 @@ def get_source_link(obj, page_info, version_tag_suffix="src/"):
     """
     Returns the link to the source code of an object on GitHub.
     """
-    package_name = page_info["package_name"]
+    repo_name = page_info["repo_name"]
```
sgugger, line 332: Can we use a get here? not sure the `page_info` will always have everything.
So try `repo_name`, then default to `package_name` if not passed.

src/doc_builder/external.py
```diff
@@ -54,14 +54,17 @@ def post_process_objects_inv(object_data, doc_url):
...
     """
     Gets the version of the last release of a package.
 
     Args:
         package_name (`str`): The name of the package.
+        repo_owner (`str`): The owner of the GitHub repo.
+        repo_name (`str`): The name of the GitHub repo. If not provided, will be the same as the package name.
     """
-    github_url = f"https://github.com/huggingface/{package_name}"
+    repo_name = repo_name or package_name
```
sgugger, line 66: Let's not rely on magic bool conversion :-)
```suggestion
    repo_name = repo_name if repo_name is not None else package_name
```

src/doc_builder/external.py
```diff
@@ -80,22 +83,25 @@ def get_stable_version(package_name):
...
     Downloads the `objects.inv` for a package and post-processes it to get a nice dictionary.
 
     Args:
         package_name (`str`): The name of the external package.
         version (`str`, *optional*, defaults to `"main"`): The version of the package for which documentation is built.
         language (`str`, *optional*, defaults to `"en"`): The langauge of the documentation being built.
+        repo_owner (`str`, *optional*, defaults to `"huggingface"`): The owner of the GitHub repo.
+        repo_name (`str`, *optional*): The name of the GitHub repo. If not provided, will be the same as the package name.
     """
+    repo_name = repo_name or package_name
```
sgugger, line 97:
```suggestion
    repo_name = repo_name if repo_name is not None else package_name
```

src/doc_builder/external.py
```diff
@@ -80,22 +83,25 @@ def get_stable_version(package_name):
...
     # We link to main in `package_name` from the main doc (or PR docs) but to the last stable release otherwise.
     if version in ["main", "master"] or version.startswith("pr_"):
         package_version = "main"
     else:
-        package_version = get_stable_version(package_name)
+        package_version = get_stable_version(package_name, repo_owner, repo_name)
 
-    doc_url = f"{HF_DOC_PREFIX}{package_name}/{package_version}/{language}"
+    doc_url = f"{HF_DOC_PREFIX}{repo_name}/{package_version}/{language}"
```
sgugger, line 104: Are we sure that we want the doc to be hosted in the `repo_name` env? I'd expect hf.co/docs/timm, not hf.co/docs/pytorch-pretrained-models

sgugger (commented):
Thanks for working on this. I think it's better to parametrize everything, just in case we end up re-using the doc-builder for other repos outside of huggingface, like bigscience (and since you have done most of the work already ;-) )

src/doc_builder/external.py
nateraw: good catch!!! Reading back through my logic here, I think this was a mistake

nateraw:
@sgugger, I _think_ we're good to go here 🤞 I made updates to reflect your review comments (thanks so much!!)

I'll need to remove hard coded references to this PR's branch in the GitHub actions before we merge, though... needed to add that so I could work on https://github.com/rwightman/pytorch-image-models/pull/1575.

.github/workflows/build_pr_documentation.yml
```diff
@@ -50,6 +50,7 @@ jobs:
         with:
           repository: 'huggingface/doc-builder'
           path: doc-builder
+          ref: repo-owner-bugfix  # TODO - remove this line before merging
```
mishig25, line 53: this is correct way to test doc-build action changes 👍

mishig25:
@nateraw, if I go to this [preview page](https://moon-ci-docs.huggingface.co/docs/timm/pr_1575/en/reference/optimizers) & click on one of the source links, for example: 
https://github.com/rwightman/pytorch-image-models/blob/vr_1575/src/timm/optim/optim_factory.py#L176

it should **not** have, string `src` & just be https://github.com/rwightman/pytorch-image-models/blob/vr_1575/timm/optim/optim_factory.py#L176

if it does not have `src` & if you replace manually `vr_1575` with `main`, the link works

sgugger (approved):
LGTM, once @mishig25 's bug is fixed of course :-)

nateraw:
@mishig25 nice catch! I added the `version_tag_suffix` build arg to the workflows so I can pass it through from `timm`. Links look good now in that preview link you shared 😄 .

mishig25 (approved):
lgtm !

PR #336, opened 2022-12-14
Add option to remove TOC check

sgugger:
This PR adds a flag in the doc config that will allow users to remove the check that all files are in the TOC. This is not activated by default because we often want to error when a contributor forgets to add a new doc file to the TOC.


LysandreJik (approved):
LGTM!

osanseviero:
Thanks a lot for the quick feature addition!

PR #339, opened 2022-12-26
Escape html vid tag

mishig25:
`<video xyz></video>` was already working since doc-builder detects opening & closing tags of html. However, when they are combined `<video xyz/>` was not being detected as HTML video tag. This PR solves it


mishig25:
Video in https://huggingface.co/docs/hub/spaces-more-ways-to-create is displyaed correctly now

Screenshot 2022-12-26 at 13 04 04

julien-c:
nice! thanks!

PR #345, opened 2023-01-25
👷 Clean-up github actions

coyotte508:
- Remove the git pull origin main
- Remove the use of the specific container, I think it was used for a secret but no longer relevant @LysandreJik ?


sgugger:
I think the container might be used to have all necessary deps of Transformers installed?

coyotte508:
ok thanks, that's helpful to know!

PR #353, opened 2023-03-02
Script to delete old docs

coyotte508:
Use `@huggingface/hub` to delete old docs, docs older than 1 month. I wanted to do three months but we're going to periodically squash the repo, and docs can be regenrated anyway.

Use Deno to run the script without the need for a `package.json`.

The Github Action will run every day.

### DIFF
--- a/scripts/delete-old-prs.ts
+++ b/scripts/delete-old-prs.ts
@@ -1,0 +1,43 @@
+#!/usr/bin/env -S deno run --allow-env --allow-net --allow-run --allow-read
+// To format: npx prettier --write .
+import { commit, listFiles } from "npm:@huggingface/hub@0.1.2";
+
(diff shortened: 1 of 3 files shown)


scripts/delete-old-prs.ts
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env -S deno run --allow-env --allow-net --allow-run --allow-read
+// To format: npx prettier --write .
+import { commit, listFiles } from "npm:@huggingface/hub@0.1.3";
```
Pierrci, line 3: Is that the main reason for picking deno? No need for `package.json` or `npm`? :)

Pierrci (commented):
👍

scripts/delete-old-prs.ts
Pierrci: (ok I should have read the description lol)

PR #359, opened 2023-03-10
Add support for building Jax notebooks

### LINKED ISSUE #358: Support Flax notebooks
Hi, it looks like the doc-builder only supports building Colab notebooks in PyTorch/TensorFlow at the moment. Would it be possible to extend this for Flax as well? For context, we're trying to build some notebooks for Diffusers, which uses Flax in some of the training doc examples.

cc @yiyixuxu @sgugger 

https://github.com/huggingface/doc-builder/blob/10fd3bbeb0eea8cff685e67c7ee414a211bde8fc/src/doc_builder/convert_to_notebook.py#L89

stevhliu:
This PR fixes #358 and it'll allow the Diffusers docs to build Jax notebooks! 🎉

### DIFF
--- a/tests/test_convert_to_notebook.py
+++ b/tests/test_convert_to_notebook.py
@@ -110,6 +110,11 @@ pt_sample
 tf_sample
 ```
 </tf>
+<jax>
+```py
+jax_sample
+```
+</jax>
 </frameworkcontent>
 End
 """
@@ -124,6 +129,9 @@ pt_sample
 ```
 ```py
 tf_sample
+```
+```py
+jax_sample
 ```
 End
 """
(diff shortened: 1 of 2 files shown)


tests/test_convert_to_notebook.py
```diff
@@ -148,6 +156,17 @@ def test_split_framewors(self):
 tf_sample
 ```
 End
+"""
+        jax_content = """
```
sgugger, line 160: Sshould be actually tested below ;-)

sgugger (commented):
Thanks for working on this! I think the test is not testing anything though.

tests/test_convert_to_notebook.py
stevhliu: D'oh! Should be tested now :)

sgugger (approved):
Thanks! Looks like it just needs a `make style` now.

PR #365, opened 2023-04-04
Don't show [lang] inference snippet if it was not supplied

mishig25:
Fixes a bug where `curl` button was showing under inference snippet even though no curl snippet was supplied in the doc file

image


coyotte508 (approved):
nice

PR #369, opened 2023-04-29
👷 Add pre_command input to CI job

### LINKED ISSUE #360: Support docs inside github artifact / url
Use case: https://github.com/huggingface/huggingface.js

There's a "docs" folder which is actually just generated docs from the source code.

It would be nice to not have to store it in the git repo, so doc-builder being able to download docs from an artificat or url would be great!

coyotte508:
Fix #360 

Slack: https://huggingface.slack.com/archives/C04U1CZQBQX/p1682775421137039 cc @xenova (hopefully it'll fix)

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -29,6 +29,9 @@ on:
         type: boolean
       install_libgl1:
         type: boolean
+      # Command to execute before building the documentation
+      pre_command:
+        type: string
       repo_owner:
         type: string
         default: 'huggingface'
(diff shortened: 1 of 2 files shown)


coyotte508:
Merging to try out in @huggingface/js

xenova:
Sweet!

coyotte508:
messed up updating build_pr_documentation - fixed in https://github.com/huggingface/doc-builder/commit/550d6e3709742a7a3366f03151d8eed210a2dc44

sgugger:
Thanks for quickly fixing the break in the PR documentation, but can we maybe wait for a review before merging something like this? It is **always** possible to test a branch by installing doc-builder from that branch or executing the workflow template from a branch, so the argument of "merging quickly to help xxx" is not a good one.

PR #370, opened 2023-04-29
Add `<include>` block for including non-code files

xenova:
Allow markdown files to include other (non-code) files. Useful to ensure minimal duplication across files.

I added a few test cases (copy-pasted and edited from `literalinclude`)

### DIFF
--- a/tests/test_convert_md_to_mdx.py
+++ b/tests/test_convert_md_to_mdx.py
@@ -145,7 +145,7 @@ export let fw: "pt" | "tf"
         text = """<include>
 {"path": "./data/convert_include_dummy.txt"}
 </include>"""
-        expected_conversion = '''<!-- START header_1 -->
+        expected_conversion = """<!-- START header_1 -->
 # This is the first header
 Other text 1
 <!-- END header_1 -->
@@ -158,7 +158,7 @@ Other text 2
 <!-- START header_3 -->
 # This is the third header
 Other text 3
-<!-- END header_3 -->'''
+<!-- END header_3 -->"""
         self.assertEqual(convert_include(text, page_info), expected_conversion)
 
         # test with indent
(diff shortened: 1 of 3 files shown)


xenova:
Not too sure what the "canonical test" is used for (defined in `literalinclude`), but I included it here too :) (perhaps for reference?)

sgugger (approved):
LGTM, but would like for @mishig25 to also have a look since he is the one who wrote that part of the code :-)

xenova:
I did run the tests on my side, and it worked correctly there. Perhaps I am using an outdated version of `black` or something :) Will check.

tests/test_convert_md_to_mdx.py
```diff
@@ -129,6 +130,62 @@ def test_process_md(self):
...
+    def test_convert_include(self):
+        path = Path(__file__).resolve()
+        page_info = {"path": path}
+
+        # test canonical
+        text = """<include>
+{"path": "./data/convert_include_dummy.txt",
+"start-after": "START header_1",
+"end-before": "END header_1"}
+</include>"""
```
mishig25, line 142: looks like this variable is overwritten (in line 145) without being used. Otherwise, the PR is lgtm !

xenova: Yep exactly what I thought when I wrote it (see https://github.com/huggingface/doc-builder/pull/370#issuecomment-1528897061)

I just copied the layout of the `literalinclude` test-case, and I assume that was just left there as a reference.

xenova: (see https://github.com/huggingface/doc-builder/blob/d563edee6ac63374dee3212b5e1a76eb7b4e426a/tests/test_convert_md_to_mdx.py#L135 for what I'm referring to)

xenova: Shall I just replace both instances with a comment?

mishig25: > Shall I just replace both instances with a comment?

Yes, please do so. And once you push the commit, happy to merge the PR 🚀

xenova: Done 👍

mishig25 (approved):
amazing work !

PR #372, opened 2023-05-12
Encode and decode `copy-to-clipboard` functionality using base64

xenova:
Fixes issue where the code attribute for `copy-to-clipboard` functionality is interpreted as valid Svelte code, which leads to SSR errors.

### DIFF
--- a/kit/preprocess.js
+++ b/kit/preprocess.js
@@ -432,6 +432,7 @@ const _mdsvexPreprocess = mdsvex({
 				lang && hljs.getLanguage(lang)
 					? hljs.highlight(lang, code, true).value
 					: hljs.highlightAuto(code).value;
+			const base64 = (val) => btoa(encodeURIComponent(val));
 			const escape = (code) =>
 				code.replace(/\\/g, "\\\\").replace(/`/g, "\\`").replace(/}/g, "\\}").replace(/\$/g, "\\$");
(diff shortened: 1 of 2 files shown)


xenova:
The failed tests seem to be related to a recent update of timm, and shouldn't have anything to do with this PR.

coyotte508:
Can you give an example of markdown that fails with current code?

xenova:
> Can you give an example of markdown that fails with current code?

Sure thing - sorry, I had mentioned to @mishig25 privately.

> # This is test\n\n
```jsx
let worker = new Worker(new URL('./worker.js', import.meta.url));
```which gives this error:

coyotte508:
if you add a newline between `#this is test` and the code sample, does the problem still happen?

xenova:
That's odd - it works locally:

image

I did notice that, when testing locally, it sometimes uses a cached version of the doc builder.

coyotte508:
Ok after a few tries  your fix worked! :) (github action cache...)

Approving, leaving it up to @mishig25 to take at second look & merge next week

coyotte508 (approved):

mishig25 (approved):
lgtm!
I've tested the build on transformers docs & everything is working as expected https://github.com/huggingface/transformers/pull/23361

Therefore, merging the PR 🚀

PR #379, opened 2023-06-02
[security] Fix exposed push tokens through gh workflow_run

mishig25:
Pretty much implements https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

tldr:
1. one workflow (without access to repo secrets) builds docs
2. another workflow (with access to repo secrets through [wokflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)) uploads/pushes the docs to the hub

todos (in order):
- [x] revert 7935dc2f6b0771f897cbdfaf0667cb579fa0c43d (done in https://github.com/huggingface/doc-builder/pull/379/commits/faac33fcb824d91d709ecc89d9b24a0df87064c0)
- [x] merge this PR
- [x] add github secret to hf organization and/or individual repos if necessary (added org level secret and messaged maintainers individually when necessary)
- [ ] update doc-build github actions of all repositories that are using doc-builder (ex: https://github.com/huggingface/transformers/pull/24013)


sgugger:
I am very uneducated in secuitry issues but what prevents the malicious user to rewrite those new workflows and get access to the secret token?

.github/workflows/delete_pr_documentation.yml
```diff
@@ -0,0 +1,72 @@
...
+    inputs:
+      package_name:
+        required: true
+        type: string
+    secrets:
+      hf_token:
+        required: true
+
+jobs:
+  upload_pr_documentation:
```
coyotte508, line 14:
```suggestion
  delete_pr_documentation:
```

.github/workflows/delete_pr_documentation.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: 'Download artifact'
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var artifacts = await github.actions.listWorkflowRunArtifacts({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: ${{github.event.workflow_run.id }},
+            });
```
coyotte508, line 46: Is it possible to get the PR number in a similar fashion?

(Currently someone can modify the workflow and overwrite the docs of another PR)

coyotte508 (approved):
Looks mostly good to me.

I think that using GH webhooks to a HF space may be better in the long run, that way no need to add the secret to each repo, which is really annoying (especially since we're going to rotate them all).

The only concern is that currently someone malicious could overwrite the docs of another PR from the same repo.

Also, we can remove the `delete_doc_comment` and `delete_pr_documentation` workflows probably: PR docs are automatically deleted after 30 days anyway.

.github/workflows/delete_pr_documentation.yml
coyotte508: For example, `${{github.event.workflow_run.pull_requests[0].number}}`

coyotte508:
> I am very uneducated in secuitry issues but what prevents the malicious user to rewrite those new workflows and get access to the secret token?

Secrets are never passed to workflows run from forks.

The workflow that uploads the docs here: https://github.com/huggingface/accelerate-wip/blob/main/.github/workflows/upload_pr_documentation.yml

It's only run from the main branch, and https://github.com/huggingface/doc-builder/pull/379/files#diff-f05826b801b9407ec985196b30fc45b111e09b5d98ee8670333493868e2b8dad  it only downloads & reuploads an artifact

.github/workflows/delete_pr_documentation.yml
mishig25: looking into it

mishig25: based on this discussion https://github.com/orgs/community/discussions/25220, it doesn't seem so
> Is it possible to get the PR number in a similar fashion?

.github/workflows/upload_pr_documentation.yml
```diff
@@ -0,0 +1,130 @@
...
+          repository: 'huggingface/doc-builder'
+          path: doc-builder
+          ref: fix-token-issue
+
+      - name: Set env variables
+        run: |
+          if [ -z "${{ secrets.token }}" ]
+          then
+            echo "comment_bot_token=$(echo 'ghp_'$(wget -qO- lysand.re/doc-build-dev)'bm')" >> $GITHUB_ENV
+          else
```
sgugger, line 40: Do we still need this hack?

sgugger (approved):
Thanks for the explanation!

Definitely worth a try (as long as it's tested with PRs opened from forks, which is the main issue we are trying to solve :-) ).

.github/workflows/upload_pr_documentation.yml
mishig25: yes, I think so. This token is for "comment bot". However, I can/will check in a separate PR if we need this token

sgugger: Yeah but this is to workaround secrets not passed. Now we can pass a secret no?

LysandreJik (approved):
If you all think it's safe, then this sounds good to me!

.github/workflows/upload_pr_documentation.yml
coyotte508: I agree we should remove all insecure secrets & rotate them

.github/workflows/delete_pr_documentation.yml
coyotte508: Did you try `${{github.event.workflow_run.pull_requests[0].number}}`?

https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run

image

It may have been added after the discussion

mishig25: https://github.com/huggingface/accelerate-wip/actions/runs/5188864703 testing [d955206](https://github.com/huggingface/doc-builder/pull/379/commits/d955206cfa72a64567cd4e02ff23fe04f84a26ee)

PR #392, opened 2023-08-12
Add support for the `pull_request_target` event in the `upload_pr_documentation` workflow

regisss:
In Optimum we need to use the `pull_request_target` event to be able to access repo secrets to build PR docs. However, PR docs have not been uploaded since this change happened.
This PR adds support for this event type in the `upload_pr_documentation` workflow.

### DIFF
--- a/.github/workflows/upload_pr_documentation.yml
+++ b/.github/workflows/upload_pr_documentation.yml
@@ -22,7 +22,8 @@ jobs:
   upload_pr_documentation:
     runs-on: ubuntu-latest
     if: >
-      github.event.workflow_run.event == 'pull_request' &&
+      (github.event.workflow_run.event == 'pull_request' ||
+      github.event.workflow_run.event == 'pull_request_target') &&
       github.event.workflow_run.conclusion == 'success'
 
     steps:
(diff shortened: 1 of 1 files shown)


sgugger:
cc-ing @coyotte508 as well as he designed the current system with @mishig25 

Just adding the workflow won't necessarily give you access to secrets, it's a tiny bit tricky and they should be able to explain it better than me.

regisss:
> cc-ing @coyotte508 as well as he designed the current system with @mishig25
> 
> Just adding the workflow won't necessarily give you access to secrets, it's a tiny bit tricky and they should be able to explain it better than me.

Actually accessing secrets to build the documentation is not the issue here, sorry I should have been more accurate. PR docs are built properly, the problem is that they are not uploaded because the "Upload PR documentation" workflow is not triggered for `pull_request_target` events.

For instance, here the PR doc was built properly: https://github.com/huggingface/optimum/actions/runs/5843309088/job/15845336669?pr=1280
But docs are never uploaded: https://github.com/huggingface/optimum/actions/workflows/upload_pr_documentation.yml

fxmarty:
@mishig25 @regisss How does it sound?

sgugger:
You should ping Mishig offline to check out this PR, he doesn't seem to read his GitHub notifications.

mishig25 (approved):
lgtm !

PR #395, opened 2023-08-29
Update README.md with missing build command

natolambert:
I think this is right. Setting up doc-builder for RLHF-handbook this time. Cheers.


mishig25 (approved):
thanks a lot!

PR #407, opened 2023-09-25
[Diffusers] Set import to slow as env variable

patrickvonplaten:
As far as I understand these environment variables will then apply to all tests, but they really only matter for the diffusers tests as the name is very diffusers specific

### DIFF
--- a/.github/workflows/build_main_documentation.yml
+++ b/.github/workflows/build_main_documentation.yml
@@ -1,5 +1,8 @@
 name: Build PR Documentation
 
+env:
+  DIFFUSERS_SLOW_IMPORT: yes
+
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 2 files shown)


mishig25 (approved):
lgtm!

.github/workflows/build_main_documentation.yml
```diff
@@ -1,5 +1,8 @@
 name: Build PR Documentation
 
+env:
+  DIFFUSERS_SLOW_IMPORT: yes
```
mishig25, line 4: double checking that `DIFFUSERS_SLOW_IMPORT=yes` means there will be no lazy loading. Correct?

patrickvonplaten: Yeah exactly

patrickvonplaten: https://github.com/huggingface/diffusers/pull/5176

patrickvonplaten: I can test then easily on this PR

PR #416, opened 2023-09-27
[Fix] trl doc build

mishig25:
### Problem

Page https://huggingface.co/docs/trl/v0.7.1/en/ddpo_trainer was [failing](https://github.com/huggingface/trl/actions/runs/6316410973/job/17169856790) with:
```
node:internal/event_target:1083
  process.nextTick(() => { throw err; });
                           ^
Error: 500 /docs/trl/main/en/ddpo_trainer
```

### Solution

Setting `sveltekit.prerender.handleHttpError` option from `fail` to `warn` solves the issue. Need a better fix. I'm not exactly sure why this error is happening at the moment. Tested the new builds locally to make sure of correctness.

image

### DIFF
--- a/kit/svelte.config.js
+++ b/kit/svelte.config.js
@@ -30,6 +30,7 @@ const config = {
 
 		prerender: {
 			crawl: false, // Do not throw if linked page doesn't exist (eg when forgetting the language prefix)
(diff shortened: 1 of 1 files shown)


mishig25:
Although https://huggingface.co/docs/trl/main/en/ddpo_trainer built fine with `500 http error`, other pages such as https://huggingface.co/docs/text-generation-inference/basic_tutorials/launcher is giving 404. Therefore, I'm gonna revert this change and work on better solution

PR #434, opened 2023-11-01
Add black as dependency at install

albertvillanova:
Add `black` as dependency at install.

Note that, besides `make quality`, `black` is also used in `style_doc.py`: https://github.com/huggingface/doc-builder/blob/e00174632acfddbf3da975dffca90a87aa334f55/src/doc_builder/style_doc.py#L21

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 
 from setuptools import find_packages, setup
 
-install_requires = ["GitPython", "tqdm", "pyyaml", "packaging", "nbformat", "huggingface_hub"]
+install_requires = ["black", "GitPython", "tqdm", "pyyaml", "packaging", "nbformat", "huggingface_hub"]
 
 extras = {}
 


mishig25:
thanks a lot!

akx:
This has apparently not been released yet?

albertvillanova:
It is true that no more official releases have been done since the last v.0.2.0 on Mar 25, 2022: https://github.com/huggingface/doc-builder/releases/tag/v0.2.0

On the other hand, we use the latest code by checking out this repository from the main branch.

Maybe some of the maintainers can give more insights...

akx:
> On the other hand, we use the latest code by checking out this repository from the main branch.

That's unfortunately not the case for Huggingface's own repositories, where I had to add an explicit hidden dependency for `black` because of this:

https://github.com/huggingface/accelerate/pull/2400
https://github.com/huggingface/peft/pull/1421

albertvillanova:
I meant in our CIs: https://github.com/huggingface/datasets/blob/f9975f636542df7f95c27065ea93147440d690b7/.github/workflows/build_documentation.yml#L13

PR #438, opened 2023-11-02
Update katex markers

mishig25:


julien-c:
Thanks, seems to be fixed indeed:

image

PR #446, opened 2023-11-17
Add `requests` test dependency

Laurent2916:
`requests` is imported here: [src/doc_builder/external.py#L22](https://github.com/huggingface/doc-builder/blob/c4e348ea1c844d659c65261d914e9cb9237a84e3/src/doc_builder/external.py#L22)


mishig25 (approved):
lgtm !

PR #454, opened 2023-11-24
Fix off-by-one error on safety condition in while loop

tomaarsen:
Hello!

## Pull Request overview
* Fix off-by-one error on safety condition in while loop, causing crash

## Details
I encountered this off-by-one error when trying to build docs for SetFit with an incorrectly formatted .mdx file. Clearly, the `idx < len(path_splits)` exists to prevent `path_splits[idx]` from failing, but `idx += 1` exists between the indexing and the checking, so we need `idx + 1 < len(path_splits)` in the check.

The docs still look good after applying the fix, but I didn't write any tests for this. I can supply tests if necessary.
You can reproduce this with e.g.:
```
[[autodoc]] TrainingArguments
  - __init__
```
(I realise that you shouldn't use `__init__`.)

- Tom Aarsen

### DIFF
--- a/src/doc_builder/autodoc.py
+++ b/src/doc_builder/autodoc.py
@@ -78,7 +78,7 @@ def get_shortest_path(obj, package):
     path_splits = long_path.split(".")
     module = package
     idx = module.__name__.count(".")
-    while idx < len(path_splits) and not hasattr(module, short_name):
+    while idx + 1 < len(path_splits) and not hasattr(module, short_name):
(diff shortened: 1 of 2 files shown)


mishig25:
Could you add testcase at https://github.com/tomaarsen/doc-builder/blob/fix/off-by-one/tests/test_autodoc.py#L169-L172 ?

mishig25 (approved):
lgtm !

PR #455, opened 2023-11-24
Allow "none" in autodoc

tomaarsen:
Hello!

## Pull Request overview
* Allow "none" in autodoc
* Add tests showing that "none" behaves like the opposite of "all"

## Details
This PR introduces another special word to autodoc beyond "all": `"none"`. Sometimes I just want it to show __init__, i.e. the default, and nothing else. This is (to my knowledge) not currently possible as providing no methods will default to using all methods.

The downside is if there's users who have a property/method etc. called `none` that they explicitly wanted to document, i.e. the exact same downside for `"all"`.

- Tom Aarsen


mishig25 (approved):
lgtm !

tomaarsen:
Awesome! Thanks for merging.

PR #461, opened 2023-12-01
Use GITHUB_OUTPUTS, not GITHUB_ENV for security reasons

mishig25:
Use `echo >> GITHUB_OUTPUT` rather than `echo >> GITHUB_ENV` because the lines below were previously security vulnerability with `echo >> GITHUB_ENV` (see hackerrank one)

https://github.com/huggingface/doc-builder/blob/c33e00834df8e85b18cdefccd56e26acb53a5be0/.github/workflows/upload_pr_documentation.yml#L75-L81

Unlikes `GITHUB_ENV`, `GITHUB_OUTPUT` gets sanitized

### DIFF
--- a/.github/workflows/upload_pr_documentation.yml
+++ b/.github/workflows/upload_pr_documentation.yml
@@ -1,121 +1,124 @@
-# name: Upload PR documentation
+name: Upload PR documentation
 
-# on:
-#   workflow_call:
-#     inputs:
-#       package_name:
-#         required: true
-#         type: string
-#       hub_base_path:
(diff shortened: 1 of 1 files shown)


mishig25:
@coyotte508 wdyt https://github.com/huggingface/doc-builder/pull/461/commits/621d1842d666887e995356fc30fccf897f170b49 ?

mishig25:
tested and working at https://github.com/huggingface/diffusers/pull/6018#issuecomment-1836126967

PR #463, opened 2023-12-05
[PR docs message] Update saying available for 30 days

mishig25:


LysandreJik:
Will the docs be automatically removed 30 days after the PR is opened or 30 days after the last update? 

The former means that if a PR stays open for longer than 30 days, we'll remove the docs before it has been merged (longest PR on transformers was maybe ~6-7 months). If the latter then I'm not sure the comment will be updated

mishig25:
> Will the docs be automatically removed 30 days after the PR is opened or 30 days after the last update?

could you comment on it @coyotte508

LysandreJik:
Perfect! So let's maybe juste update the message to say "30 days after the last update" @mishig25 ?

PR #466, opened 2024-01-17
Add `workflow_dispatch` to events that trigger upload PR doc

regisss:
In Optimum, we sometimes need to trigger the PR doc build manually: when modifying the [PR doc build workflow](https://github.com/huggingface/optimum/blob/main/.github/workflows/build_pr_documentation.yml) in a PR, the only way to execute the PR version of this workflow (and not the one from `main`) is with a `workflow_dispatch` event. This happens because we use the `pull_request_target` event to be able to access secrets during doc generation.

This PR adds the `workflow_dispatch` event as a compatible event for the doc upload workflow.


mishig25 (approved):
lgtm !

PR #469, opened 2024-01-29
Better support for notebooks-to-md

mishig25:
Extend functionalities to #263 
- [x] support converting all notebooks in a folder (previously, you'd need to specify specific notebook file path)
- [x] add notebook conversion support in github workflows
- [x] fix escpaing problems by using html `<pre>` when necessary
- [ ] support img/audio outputs (will be in separate PR)

### DIFF
--- a/src/doc_builder/commands/notebook_to_mdx.py
+++ b/src/doc_builder/commands/notebook_to_mdx.py
@@ -14,29 +14,59 @@
 # limitations under the License.
 
 import argparse
-import subprocess
 from pathlib import Path
 
+import nbformat
 from tqdm import tqdm
+
+from ..style_doc import format_code_example
+
+
+def notebook_to_mdx(notebook, max_len):
+    content = []
+    for cell in notebook["cells"]:
+        if cell["cell_type"] == "code":
+            code = cell["source"]
+            outputs = [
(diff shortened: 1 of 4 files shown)


mishig25:
lgtm ✅

src/doc_builder/commands/notebook_to_mdx.py
```diff
@@ -14,60 +14,44 @@
...
-        f.write("\n\n".join(content))
+    src_path = Path(args.notebook_src).resolve()
+    src_dir = src_path.parent if src_path.is_file() else src_path
+    notebook_paths = [str(src_path)] if src_path.is_file() else [*src_dir.glob("*.ipynb")]
+    output_dir = src_dir if args.output_dir is None else Path(args.output_dir).resolve()
+
+    try:
+        for notebook_path in tqdm(notebook_paths, desc="Converting .ipynb files to .md files"):
+            subprocess.run(
+                ["jupyter", "nbconvert", notebook_path, "--to", "markdown", "--output-dir", output_dir],
```
julien-c, line 32: not sure if it makes sense/if it's relevant but just noting that we have an internal nbconvert API here: https://huggingface.co/spaces/julien-c/nbconvert

mishig25: good to know. But right now I think it is fine to use `nbconvert` locally rather than through API

PR #483, opened 2024-03-07
Add release directions + license to setup.py

muellerzr:
Adds release directions to the setup.py to mimic the rest of our O.S. libraries

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,3 +31,31 @@ setup(
     install_requires=install_requires,
     entry_points={"console_scripts": ["doc-builder=doc_builder.commands.doc_builder_cli:main"]},
 )
+
+# Release checklist
+# 1. Checkout the release branch (for a patch the current release branch, for a new minor version, create one):
+#      git checkout -b vXX.xx-release
+#    The -b is only necessary for creation (so remove it when doing a patch)
+# 2. Change the version in __init__.py and setup.py to the proper value.
+# 3. Commit these changes with the message: "Release: v<VERSION>"
+# 4. Add a tag in git to mark the release:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,6 +19,7 @@
     name="hf-doc-builder",
     version="0.5.0.dev0",
     author="Hugging Face, Inc.",
+    license="Apache",
     author_email="sylvain@huggingface.co",
```
muellerzr, line 23: @LysandreJik who should we put for this now?

setup.py
```diff
@@ -31,3 +32,31 @@
...
+# 4. Add a tag in git to mark the release:
+#      git tag v<VERSION> -m 'Adds tag v<VERSION> for pypi'
+#    Push the tag and release commit to git: git push --tags origin vXX.xx-release
+# 5. Run the following commands in the top-level directory:
+#      rm -rf dist
+#      rm -rf build
+#      python setup.py bdist_wheel
+#      python setup.py sdist
+# 6. Upload the package to the pypi test server first:
+#      twine upload dist/* -r testpypi
```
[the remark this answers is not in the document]
muellerzr: I don't think we've considered doing that honestly. We could make a `release` dep section for it I suppose? Since its only ever needed here. But it's only for core maintainers, users should never, ever, need to do this and it's only done once a month (or less) when a release is made.

setup.py
LysandreJik: mishig's email? wdyt @mishig25 ?

mishig25: docs@huggingface.co would be good

mishig25 (approved):
lgtm !

PR #485, opened 2024-03-15
attempt at replacing black ith ruff

sayakpaul:


src/doc_builder/style_doc.py
```diff
@@ -91,6 +91,7 @@ def parse_code_example(code_lines):
     return code_samples, outputs
 
 
+# Code relies on black API.
```
sayakpaul, line 94: This function implicitly relies on `black`. We need to find a way to replace it. `ruff` doesn't have any API. 

```
def format_code_example(code: str, max_len: int, in_docstring: bool = False):
    """
    Prepare a code example for formatting. This version is adapted for use without direct automatic formatting like 'black'.
    The function prepares code by handling initial indentation, doctest syntax, and docstring specifics.

    Args:
        code (str): The code example to format.
        max_len (int): The maximum length per line, not used directly in formatting here but for your reference.
        in_docstring (bool, optional, defaults to False): Whether the code example is inside a docstring.

    Returns:
        str: The prepared code, not automatically formatted.
    """
    code_lines = code.split("\n")

    # Find initial indent
    idx = 0
    while idx < len(code_lines) and is_empty_line(code_lines[idx]):
        idx += 1
    if idx >= len(code_lines):
        return "", ""
    indent = find_indent(code_lines[idx])

    # Remove the initial indent for now, we will add it back later.
    code_lines = [l[indent:] for l in code_lines[idx:]]
    has_doctest = code_lines[0][:3] in DOCTEST_PROMPTS

    code_samples, outputs = parse_code_example(code_lines)

    formatted_lines = []
    for code_sample, output in zip(code_samples, outputs):
        # Process code samples for doctest
        code_sample = code_sample.strip()
        in_triple_quotes = False
        in_decorator = False
        for line in code_sample.strip().split("\n"):
            if has_doctest and not is_empty_line(line):
                prefix = (
                    "... "
                    if line.startswith(" ") or line[0] in [")", "]", "}"] or in_triple_quotes or in_decorator
                    else ">>> "
                )
            else:
                prefix = ""
            indent_str = "" if is_empty_line(line) else (" " * indent)
            formatted_lines.append(indent_str + prefix + line)

            if '"""' in line:
                in_triple_quotes = not in_triple_quotes
            if line.startswith(" "):
                in_decorator = False
            if line.startswith("@"):
                in_decorator = True

        formatted_lines.extend([" " * indent + line for line in output.split("\n")])
        if not output.endswith("===PT-TF-SPLIT==="):
            formatted_lines.append("")

    result = "\n".join(formatted_lines)
    return result.rstrip(), ""

```

Is what ChatGPT produced.

mishig25:
Any idea why the tests are failing ?

sayakpaul:
Fixed.

mishig25:
Tests are still failing. I was in assumption that this PR would only be one-line change
```diff
- black
+ ruff
``` 

I don't think it is worth your time for this change that is requiring more work than the value

sayakpaul:
Certainly not a one line change :3

PR #533, opened 2024-10-16
Pin ubuntu-22.04 in the build_main_documentation workflow

regisss:
See https://huggingface.slack.com/archives/C1RCG46LS/p1729118325934509?thread_ts=1728616969.079469&cid=C1RCG46LS.


regisss:
@mishig25 It's probably safer to pin ubuntu-22.04 for now

PR #541, opened 2025-01-16
bump `actions/upload-artifact` version

gante:
`actions/upload-artifact@v3` is no longer working, see e.g. [this workflow](https://github.com/huggingface/transformers/actions/runs/12813495878/job/35727693015?pr=35709)

PR #544, opened 2025-01-24
Fix PR doc upload workflow

regisss:
`actions/upload-artifact@v3` recently had to be upgraded because it is now deprecated (see https://github.com/huggingface/doc-builder/pull/541).
However, the logic used to download the uploaded artifact does not work anymore with this upgrade. This PR fixes this issue and replaces the custom download logic with `actions/download-artifact@v4`.


glegendre01 (approved):
Sounds good

echarlaix:
Thanks a lot @regisss!

PR #546, opened 2025-03-20
[CI] update PR doc builder machine

gante:
Fix for doc builder failure due to lack of disk space in `transformers`, e.g. https://github.com/huggingface/transformers/actions/runs/13968457454/job/39104194581?pr=36850


ydshieh (approved):
thanks.
we might need the same for the build doc. on main

PR #547, opened 2025-03-20
[CI] Optional runner arg

gante:


gante:
(closing: transformers found a solution that does not imply using groups, i.e. base GHA images are used)

PR #601, opened 2025-07-18
[Tip] Add icons

stevhliu:
Adds a "warning" and "lightbulb" icon for the Warning/Tip component as shown below. This helps add some more context to the current Tip component. I can also explicitly add a "Warning" and "Tip" label to the component if we like :)

Screenshot 2025-07-18 at 1 23 22 PM


gary149:
spacing looks a bit too large to me @enzostvs + how does it look with a single sentence going on multiple lines.

enzostvs:
Agreed, the icon isn't aligned with the text as well and too much space on the right, maybe could be 1,5 instead of 2,5

stevhliu:
Changed the icon to be more aligned and wrapped the text under the first line!

Screenshot 2025-10-16 at 3 14 10 PM

mishig25:
awesome, I will do some local testing and merge 🚀

stevhliu:
thanks all! 🙏

PR #615, opened 2025-08-12
fix checkout specific commit

ydshieh:
Let's avoid issue by only running checkout if event is issue_comment

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -80,6 +80,7 @@ jobs:
           path: ${{ inputs.package }}
 
       - name: Checkout to the correct commit
+        if: github.event_name == 'issue_comment'
         working-directory: ${{ inputs.package }}
         run: |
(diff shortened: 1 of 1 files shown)


gante (approved):
LGTM 👍

.github/workflows/build_pr_documentation.yml
```diff
@@ -80,6 +80,7 @@ jobs:
           path: ${{ inputs.package }}
 
       - name: Checkout to the correct commit
+        if: github.event_name == 'issue_comment'
```
gante, line 83:
```suggestion
        if: ${{ github.event_name == 'issue_comment' }}
```

(from the example in this part of the docs: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#example-usage-of-the-github-context)

ydshieh: without it , still works, but good to have this format

PR #626, opened 2025-09-12
Fix Latex example in readme

burtenshaw:
This PR aligns the docs on latex syntax with the functionality based on the example [here](https://github.com/huggingface/transformers/blob/main/docs/source/en/model_doc/rwkv.md?plain=1#L93)


qgallouedec:
For some reason, you actually need **2 whitespaces** before the inline latex, otherwise it renders like this.

It's a but from this packages, but I was not able to fix it, so I share a workaround in case it's useful 

Screenshot 2025-09-17 at 3 59 06 PM

PR #639, opened 2025-09-27
Support dollar math inline syntax

cyyever:


cyyever:
@mishig25 I have added.

cyyever:
@mishig25 The current failure looks unrelated.

PR #643, opened 2025-09-30
feat: add github app authentication instead of comment_bot_token

paulinebm:
Hi, 

We have to change the way we inject COMMENT_BOT_TOKEN. Indeed, we have reached the number of repositories able to access this fine-grained token. 

I created a GitHub app to bypass this limitation. We also improve security by generating a temporary token for each workflow execution. This token is revoked automatically at the end of the process. 

I introduce 2 news secrets to achieve that. 

Could you just approve, I will merge at the end of the workflows update task. 

Thanks

cc @glegendre01 @jagwar for viz

### DIFF
--- a/.github/workflows/upload_pr_documentation.yml
+++ b/.github/workflows/upload_pr_documentation.yml
@@ -15,8 +15,12 @@ on:
     secrets:
       hf_token:
         required: true
-      comment_bot_token:
-        required: true
+      comment_bot_token: #legacy, will be removed in the future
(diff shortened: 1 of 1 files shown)


glegendre01 (approved):
LGTM

Wauplin (approved):
Thanks for the ping! Looks good to me 👍

Wauplin:
(failing tests are unrelated btw, I haven't got the time to look into them)

PR #787, opened 2026-05-05
feat: add language-* class to rendered code blocks

mishig25:
## Summary
- The fence language identifier (e.g. `python` in ` ```python `) was parsed by mdsvex and used only to pick the syntax highlighter, then discarded before rendering.
- Forward it through `<CodeBlock>` / `<CodeBlockFw>` and emit `class=\"language-<lang>\"` on the `<pre>`, so consumers can target code blocks by language from CSS or external tooling.
- Sanitized via `/^[\w-]+$/`; fences without a language (or with non-identifier characters) get no class, preserving prior behavior.


stevhliu (approved):
thanks!

PR #813, opened 2026-08-13
Add a translation pipeline

stevhliu:
This PR adds a pipeline for translating the English Transformer docs into another language. The English files in `docs/source/en` remain the source of truth.

Each run reads a specific Transformers commit, including all docs pages and the sidebar. It protects code, links, doc-builder syntax, API references, and technical names before sending the remaining prose to the model. A glossary controls names and labels that should stay in English. A CPU-only preflight can check syntax preservation before spending GPU time.

The pipeline reuses complete translated pages when their source and translation settings still match the cache. For the remaining pages, a Job runs Qwen3 using Transformers continuous batching. If every page is cached, generation is skipped and the model does not load.

Translations are checked for missing or changed protected content and document structure. Rejected passages get one retry, using smaller pieces where needed. Only completed pages that pass validation enter the cache. If failures remain, the run does not produce a build artifact.

Completed translations are stored in the `hf-doc-build/doc-translate` Bucket. The Markdown is readable directly under `transformers/[language-code]`. Each run also stores its own source files, archive, and metadata. The full-run coordinator verifies the results and updates the shared translation cache.

The shared doc-builder workflow can download a specific completed archive, verify it, and install it as the other language docs source. It then builds the docs using the existing HTML page cache.

## Preview

From your doc-builder repo, preview with:

```bash
source /tmp/ja-preview-env/bin/activate

hf buckets sync \
  hf://buckets/hf-doc-build/doc-translate/transformers/ja \
  /tmp/transformers-ja-preview \
  --exclude '.translation-state.json'

doc-builder preview transformers /tmp/transformers-ja-preview \
  --language ja --version main
```


coyotte508 (commented):

This PR adds a `doc-builder translate` command — a nightly, cache-aware machine-translation pipeline (mask → segment → translate via continuous batching → validate → publish to a bucket) plus a sync step in the shared build workflow to swap translated pages into the docs build. The design is thoughtful, the comments explain real production incidents, and the non-GPU logic has genuinely strong test coverage; no outright bugs found, but several gaps worth addressing.

**Findings**

- 🟡 **Deleted English pages are never pruned from the bucket** — `run()` in `src/doc_builder/commands/translate.py` only writes pages (`write_page`); nothing removes files from `translations/<pkg>/<lang>` when the English original disappears. The workflow comment ("clears the 20 stale orphans") only holds for repo-side orphans: future orphans accumulate in the bucket, get synced into `docs/source/<lang>` wholesale, and are published unlisted forever. Consider deleting out-dir files not in the current page set.
- 🟡 **A `--pages-file` smoke run against the real bucket clobbers production** — `run()` writes the *pruned* `_toctree.yml` (via `prune_toctree`) to the same `out_dir` a full run uses. Someone pointing a 3-page smoke run at the production bucket replaces the full sidebar with a 3-entry one until the next full run. Guard it (e.g. refuse `--pages-file` without an explicit non-default `--bucket`, or write to a scratch prefix).
- 🟡 **Exit code ignores page-level failure rate** — the command's own docstring says "the exit code is the only way anyone finds out something went wrong," yet `run()` returns 0 even if e.g. 90% of pages fail validation and fall back to English; `validate.summarize()` only *prints* a warning above 2%. Consider a non-zero exit (or configurable threshold) so the nightly job goes red.
- 🟡 **Unpinned `transformers` in the `translate` extra** — `pipeline.translate_segments()` depends on v5 continuous-batching internals (`ContinuousBatchingConfig`, `continuous_batching_context_manager`, `transformers.generation.continuous_batching.utils.WorkloadHints` — a private-ish import path), but the extra declares bare `"transformers"`. Pin a minimum version; `kernels` got one, `transformers` deserves one more.
- 🟡 **Glossary `keep:` terms never reach the model** — `build_prompt()` only injects `pin` entries; `keep` (ja.yml: "Words to leave in English") is only checked *after* the fact as a warning in `check_glossary()`. If that's intentional (rely on masking + warning), the comment in `glossaries/ja.yml` oversells what `keep` does; otherwise add them to the prompt.
- 🟡 **Language-list format mismatch in the workflow** — the existing `languages` input is parsed with `IFS=', '` (commas allowed, `build_main_documentation.yml` L211), but the new sync loop does plain word-splitting on `$TRANSLATED_LANGUAGES`. A caller who copies `"ja, ko"` into both inputs gets a bogus `ja,` bucket path (it fails loudly at the `_toctree.yml` check, at least). Normalize commas, and consider enforcing the documented "must also appear in `languages`" invariant instead of just stating it.
- 🟡 **The "load-bearing" tests never run in CI** — `test_translate_segment.py` / `test_translate_pipeline.py` call the corpus round-trip "the load-bearing test in this project," but `EN_DOCS` defaults to a personal path (`/Users/steven/hf/transformers/docs/source/en`) and skips everywhere else. Vendor a small fixture corpus (a dozen representative pages) so a slice of the round-trip runs on every CI run, and drop the personal path default.
- 🔵 Missing Apache license headers on `translate/cache.py`, `translate/pipeline.py`, `translate/validate.py`, and all four new test files — `segment.py`, `translate/__init__.py`, and `commands/translate.py` have them, and every existing module in the repo carries one.
- 🔵 `select_pages()` filters comments with `line.startswith("#")` on the *unstripped* line, so an indented `  # comment` in a pages file becomes a "missing page" warning.
- 🔵 `LANGUAGE_NAMES` only knows `"ja"`; any other `--lang` silently injects the raw code into the prompt ("from English into ko"), which will degrade quality with no warning.
- 🔵 A swapped link pair (`¤1¤label¤0¤`) passes the per-paragraph *sorted* marker check in `assemble_page()` and is only caught page-level by `check_links()` — so one such paragraph rejects the whole page, defeating the per-paragraph containment the code works hard for elsewhere. Checking pair order for `link_open`/`link_close` markers per paragraph would keep the blast radius small.
- 🔵 `run()` has no per-page try/except around `assemble_page`/`validate_plan`; an unexpected exception on one page after GPU spend aborts the write-out of every remaining page — at odds with the "one bad page costs itself" philosophy.
- 🔵 `test_translate_pipeline.py` module docstring references `plan_page`, which doesn't exist (it's `PagePlan`); the new workflow step also omits `shell: bash` where sibling steps declare it.

**Questions:** is `google/gemma-4-26B-A4B-it` gated, and does the job's `HF_TOKEN` have access? Does `hf sync` (bucket → staged dir) delete extraneous destination files, i.e. does the workflow-side swap actually mirror? Is `os.replace()` atomic on the bucket FUSE mount that `SegmentCache` relies on for its tmp-then-rename writes?

**Verdict:** Well-engineered and unusually well-tested for a pipeline of this kind — no blockers, but the orphan-cleanup, smoke-run-toctree, and exit-code items should be resolved (or explicitly deferred) before merge.

src/doc_builder/translate/segment.py
```diff
@@ -0,0 +1,205 @@
...
+#   elements split over two lines, and if we left those alone their URLs would be handed
+#   to the model as if they were prose. We checked -- allowing this picks up those 88 and
+#   nothing else. The longest real tag is 326 characters, so the 600 limit is just a
+#   backstop for a malformed tag that never closes.
+#
+#   "code" is not allowed to cross a line break, so one stray backtick cannot hide the
+#   rest of the document.
+MASK_PATTERNS = [
+    ("comment", re.compile(r"<!--.*?-->", re.DOTALL)),
+    ("fence", re.compile(r"^[ \t]*(`{3,}|~{3,}).*?^[ \t]*\1[ \t]*$", re.DOTALL | re.MULTILINE)),
```
mishig25: [P1] This backreference requires the closing fence to have exactly the opener's length, but CommonMark allows a closing fence with at least as many delimiters. In the real Transformers `model_doc/mms.md`, a three-backtick `bash` fence closes with four backticks. `mask()` skips that valid closer and consumes the following English paragraph plus the next Python block through a later triple fence, so that prose is never translated and validation cannot see it. Please parse the delimiter character and minimum length (or use a Markdown parser), and add this corpus case as an over-masking test.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,357 @@
...
+    )
+
+    # This is the moment the whole design is built around: decide whether there is anything to
+    # do before going anywhere near the model.
+    #
+    # --rebuild carries on anyway, using the cache. Nothing here needs a GPU when there is
+    # nothing to translate, so it is a cheap way to republish every page after a change to how
+    # pages are assembled or checked. Without it those changes never reach the bucket, because
+    # a warm run stops before it ever rebuilds a page.
+    if not missing and not args.rebuild:
```
mishig25: [P1] `missing` only answers whether the model has new prose to translate; it does not tell us whether the source tree or published output changed. A code/URL/tag-only edit keeps the same masked segment key, and deleting or reordering existing paragraphs can also leave `missing` empty, so this return preserves stale output. It also prevents a warm cache from repairing a missing/partial translation tree. I reproduced both a masked-only edit and a paragraph deletion remaining stale. Please separate the model-load decision from output reconciliation, using a source/output manifest or rebuilding affected pages from cache.

src/doc_builder/translate/pipeline.py
```diff
@@ -0,0 +1,511 @@
...
+def build_prompt(segment_text, language, glossary):
+    """Build the instructions and the paragraph into a chat message for the model."""
+    terms = glossary_for_segment(segment_text, glossary)
+    if terms:
+        lines = "".join(f"\n  - {t} -> {r}" for t, r in sorted(terms.items()))
+        glossary_block = GLOSSARY_HEADER + lines
+    else:
+        glossary_block = ""
+
+    system = SYSTEM_PROMPT.format(
```
mishig25, line 193: [P1] The system prompt is formatted with `⟦0⟧`, while `segment.mask()` actually emits `¤0¤`. The preservation rule therefore describes tokens that are absent from the user input and gives no explicit instruction for the real placeholders; it also likely explains the echoed `⟦N⟧` markers being cleaned up later. Please use `segment.PH_OPEN`/`PH_CLOSE` here and bump `PROMPT_VERSION` so cached results produced under the mismatched prompt are invalidated.

src/doc_builder/translate/pipeline.py
```diff
@@ -0,0 +1,511 @@
...
+    """
+    parts = list(plan.parts)
+    rejected = []
+    for index, unit in plan.units.items():
+        translated = translations.get(unit.key)
+        if translated is None:
+            continue
+        # Cleaned on the way out rather than on the way in, so a cache full of translations
+        # written before this existed is fixed by a --rebuild, with nothing retranslated.
+        translated = strip_echoed_markers(translated)
```
mishig25: [P1] Empty or marker-only model output is accepted here. For a prose unit with no placeholders, `translated == ""` passes both checks and replaces the paragraph with nothing; if another unit translated successfully, whole-page validation still passes. I reproduced a passing page with one body paragraph silently removed, and empty toctree titles have the same issue. Reject translations whose non-placeholder prose is empty before caching and again during assembly.

src/doc_builder/translate/segment.py
```diff
@@ -0,0 +1,205 @@
...
+    #
+    # Now they are, so a mangled link is just a missing marker, and the ordinary check catches
+    # it with no special case. `link_open` only fires on a `[` that actually starts a link,
+    # which is what the lookahead is for.
+    # The lookahead allows one level of brackets inside the label, because 144 links in the docs
+    # have them: `[![Open In Colab](img)](link)` wraps an image in a link, and
+    # `[huggingface_hub[cli]](url)` has them in the text. A simpler lookahead that forbade any
+    # `]` before the `](` skipped those, so only the closing bracket got hidden and the exposed
+    # `[` came back -- along with the model's habit of "closing" it.
+    ("link_open", re.compile(r"\[(?=(?:[^\[\]\n]|\[[^\[\]\n]*\])*\]\([^)\n]*\))")),
```
mishig25: [P1] The link lookahead/closer stops the destination at the first `)`, so balanced parentheses expose the URL suffix. For example, `[LED](https://e/Fine_(LED)_guide.ipynb)` masks as `¤0¤LED¤1¤_guide.ipynb)`. The model can alter that suffix and `check_links()` still reports success because its regex makes the same partial match. This occurs in the real corpus. Please scan balanced Markdown destinations (and reference-style links) and compare exact destination/type multisets after restoration rather than only counting partial link matches.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,357 @@
...
+    results, written, skipped = [], 0, 0
+    for page, plan in plans.items():
+        masked_translation, page_text, rejected = pipeline.assemble_page(plan, available)
+        result = pipeline.validate_plan(plan, masked_translation, glossary, page_text)
+        if rejected:
+            result.warnings.append(f"{len(rejected)} paragraph(s) kept in English: markers not preserved")
+        results.append(result)
+        if result.ok:
+            disclosed = pipeline.add_disclosure(page_text, page, args.lang, args.package)
+            write_page(out_dir, page, disclosed)
```
mishig25: [P1] Pages are written directly into the live bucket after the cache has already been committed. The old `_toctree.yml` remains present throughout, so an interrupted or concurrent run can leave a mixed tree that the workflow's existence check accepts; the next warm run may then skip repairing it. Removed pages are also never pruned, and a 100% translation failure currently writes English fallbacks and returns success. Please assemble a complete staged/versioned tree, enforce a catastrophic-failure threshold, validate it, and only then promote it with deletion semantics.

mishig25 (changes requested):
The overall design is promising, but the inline findings below identify cases that can silently publish stale, missing, or corrupted documentation. I reproduced the warm-cache stale-output path, silent paragraph deletion, and successful exit after total translation failure; corpus testing exposed the fence and link cases. Please address these before enabling the nightly pipeline.

Local verification: 1,687 tests passed with the current Transformers corpus, and lint/formatting passed. `uv lock --check` currently fails. The actual GPU model and bucket-to-build path were not exercised. Follow-ups worth covering include `$...$` math, reference/bare URLs and image structure, safe `--pages-file` output, `.mdx` support, glossary `keep:` behavior, a minimum Transformers version plus regenerated lockfile, and running representative corpus tests in CI.

stevhliu:
Thanks, I pinned `transformers >=5.8.0` and regenerated `uv.lock`! Here's a brief summary of the requested changes:

- Updated the masking logic to accommodate more backticks, mask link URLs with parentheses in them, and mask reference-style links as well (`[text][ref]`)
- Updated the prompt (used `⟦0⟧` but the text actually uses `¤0¤`) to use the same markers the text actually contains.
- Refuse empty translations which used to pass the checks.
- Link check compares the actual URLs instead of counting links.
- Updated the publishing workflow as well:
   - build and check the docs in memory first before writing to bucket
   - docs whose original English source was deleted are removed from the bucket
   - new manifest that tracks what each translated doc is built from, which is what lets the Job detect a code-only edit, a deleted paragraph, etc. (none of which the segment cache could see before)
   - if more than 25% pages fail their checks, nothing is published and the Job exits with an error
   - if less than 2% pages fail, the docs are published but it exits with a non-zero value on the nightly Job to flag a look
- Added real docs to `tests/fixtures/translate_corpus/` so the masking tests run in CI

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,467 @@
...
+    pages are shortcuts to files outside the docs folder, so if one of those ever points
+    somewhere we did not fetch, it should cost us that page and not the whole night.
+    """
+    plans, unreadable = {}, []
+    for page in pages:
+        try:
+            text = (source_dir / page).read_text(encoding="utf-8")
+        except OSError as exc:
+            unreadable.append(f"{page} ({exc.__class__.__name__})")
+            continue
```
mishig25, line 103: [P1] Fail closed on incomplete source snapshots. This skip is now destructive: the omitted page is absent from `sources`, so `publish_tree()` later treats its live translation as an orphan. I reproduced one transient `OSError` for `guide.md`; the run returned 0, deleted the published page, and left `_toctree.yml` still referencing it. With an empty or wrong source directory, the same path removes the entire live tree. A full publish should require a readable sidebar and a complete/nonempty expected page set, then abort or explicitly preserve skipped pages on any read failure rather than treating them as deletions.

src/doc_builder/translate/publish.py
```diff
@@ -0,0 +1,237 @@
...
+    written = 0
+    for page, text in sorted(tree.items()):
+        dest = out_dir / page
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            if dest.is_file() and dest.read_text(encoding="utf-8") == text:
+                continue
+        except OSError:
+            pass
+        dest.write_text(text, encoding="utf-8")
```
mishig25, line 315: [P1] Publish a complete generation atomically. This still mutates files in the live directory one at a time, and sorting means `_toctree.yml` is normally written first. I simulated the next write raising and the visible tree contained the new sidebar alongside old or missing pages. Writing the manifest last only lets a later run repair the damage; the build workflow does not consult that manifest, so a concurrent sync can publish the mixture. Please write an immutable/staged generation and expose it only through an atomic promotion or completion pointer, with failed removals treated as publication failure.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,467 @@
...
+    the check that matters, since it is the one that has caught pages published under older,
+    weaker rules.
+    """
+    if not path.is_file():
+        return None
+    try:
+        published = path.read_text(encoding="utf-8")
+    except OSError:
+        return None
+    problems = validate.check_links(source, published)
```
mishig25: [P1] Account for generated disclosure links when retaining a page. Every successful page has the English-original and issue-report links injected before publication, but this compares that augmented page with the raw English source. As a result, `check_links()` always reports those two links as gained, so every page produced by this command is rejected as the last-good fallback. I reproduced a failed update replacing a valid Japanese page with raw English, after which the manifest marked that fallback current. Validate against a banner-aware expected source or remove/parse the generated disclosure before comparing links.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,467 @@
...
+        tree[TOCTREE] = toctree_text
+
+    print(validate.summarize(results))
+
+    # Refuse to publish a run that went badly wrong. Every failed page falls back to English or
+    # to whatever it had before, so a night where the model returns rubbish for everything used
+    # to look like a success: 732 English pages written into the Japanese docs, exit code 0.
+    # Leaving the previous translation in place is better than that, and it means one bad
+    # night cannot undo a good one.
+    rate = len(failed) / len(results) if results else 0.0
```
mishig25, line 483: [P1] Include request failures and English fallbacks in the failure gate. `rate` counts only whole-page validation failures; missing translations simply remain English, and rejected paragraphs only add warnings. I reproduced 4 of 5 model requests failing: all three body paragraphs stayed English, one translated heading made the page non-identical to its source, validation reported 1/1 pages passed with a 0.0% rejection rate, and the run published and returned 0. Please gate on attempted-segment success/translated coverage, or mark pages below a minimum coverage as failed, including sidebar-title failures.

src/doc_builder/translate/validate.py
```diff
@@ -0,0 +1,264 @@
...
+from .segment import LINK_DEST, PLACEHOLDER_RE, REF_DEF_RE, placeholder_indices
+
+HEADING_RE = re.compile(r"^(#{1,6})[ \t]+\S", re.MULTILINE)
+
+# A markdown link or image, e.g. [text](url), with its destination captured. It deliberately
+# will not match across a line break -- that restriction is what lets us notice a marker that
+# has drifted onto another line. The destination pattern is shared with segment.py so the two
+# cannot drift apart: when they did, a URL with brackets in it matched partially on both sides,
+# the counts agreed, and a rewritten URL sailed through.
+MD_LINK_RE = re.compile(rf"(?P<image>!?)\[[^\]\n]*\](?P<dest>{LINK_DEST})")
```
mishig25: [P1] Preserve and parse nested image links. This regex stops at the first `]`, so for `[Open In Colab](notebook.ipynb)` it records only `badge.svg` as an ordinary link and never records the outer destination or image structure. The masker also leaves `!` exposed. Removing that `!` preserves every numbered placeholder and produces `[[Open In Colab](badge.svg)](notebook.ipynb)`, yet whole-page validation still succeeds. This shape occurs throughout the corpus. Mask the image opener as protected syntax and use a parser or shared nested-label grammar that validates both destinations and link/image kinds.

src/doc_builder/translate/segment.py
```diff
@@ -0,0 +1,281 @@
...
+    # down the page. Only four of these exist in the docs, all in `model_doc/gemma3n.md`, and
+    # none of them were masked at all -- so the label `altup` went to the model as prose, where
+    # translating it would break the link silently.
+    #
+    # Handled the same way as an ordinary link: both brackets and the reference become markers,
+    # the visible text between them stays translatable.
+    ("ref_link", _ref_link_patterns),
+    # The definitions themselves go whole. There is nothing to translate in `[altup]: https://…`
+    # -- the label is an identifier and the rest is a URL.
+    ("ref_def", REF_DEF_RE),
```
mishig25, line 280: [P1] Mask bare URLs as protected content. The masking list protects Markdown and reference destinations but has no rule for raw URLs, and `check_links()` does not compare them. In the included `deepseek_v3.md` fixture I changed the exposed `https://github.com/deepseek-ai/DeepSeek-V3` destination; assembly rejected nothing and whole-page validation returned OK. Raw URL-bearing lines occur across the current corpus, while this module promises that URLs remain exact. Please mask raw URLs/autolinks and compare them after restoration.

src/doc_builder/translate/segment.py
```diff
@@ -0,0 +1,281 @@
...
+        ),
+    ),
+    # Matches what build_doc actually accepts (see _re_autodoc and _re_list_item there):
+    # the directive may be indented, and blank lines may sit before its list of methods.
+    # An earlier version only matched it at the start of a line, which missed the indented
+    # one in the docs and sent `[[autodoc]] LevitImageProcessor` off to be translated.
+    ("autodoc", re.compile(r"^[ \t]*\[\[autodoc\]\][^\n]*(?:(?:\n[ \t]*)*\n[ \t]*-[ \t]+[^\n]*)*", re.MULTILINE)),
+    ("directive", re.compile(r"\[\[[^\]\n]*\]\]")),
+    ("math_block", re.compile(r"\$\$.*?\$\$", re.DOTALL)),
+    ("math_inline", re.compile(r"\\\\\(.*?\\\\\)", re.DOTALL)),
```
mishig25, line 194: [P1] Mask standard inline math. These entries cover `$$...$$` and the escaped parenthesis form, but ordinary `$...$` formulas remain model-visible. I reproduced changing `$K_{\\text{past}}$` to a corrupted formula with no rejected paragraph and a successful page validation; this form occurs repeatedly in the corpus. Add a tokenizer or carefully bounded inline-dollar rule, with fixtures that distinguish formulas from literal currency.

mishig25 (changes requested):
Thanks for the follow-up. The fence handling, source/output reconciliation, prompt marker fix, empty-output guard, balanced destinations, and full reference-link support are all materially improved. The focused translation suite passes (123 passed, 5 skipped), and the current GitHub checks are green. However, the inline cases below can still delete or expose an incomplete live tree, turn a failed update into a successful English fallback, or silently corrupt protected content. I reproduced each case against this head, so I am keeping the review at changes requested before the nightly pipeline is enabled.

mishig25:
could you resolve the conflicts? 

image

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+            print(
+                f"[translate] ERROR {len(unreadable)} page(s) could not be read: {unreadable[:5]}; "
+                "refusing to publish from an incomplete source"
+            )
+            return 2
+        print(f"[translate] WARNING skipped {len(unreadable)} unreadable page(s): {unreadable[:5]}")
+    # If we are only translating some pages, the sidebar has to be trimmed to match, or the
+    # result cannot be built.
+    subset = pages if preview else None
+    toc_tree, toc_keys, toctree_text = load_toctree(source_dir, args.lang, args.model, gloss_sha, subset)
```
mishig25, line 351: [P1] Reject sidebar entries whose source page is missing. A full run still derives `pages` only from files that happen to exist and never checks those pages against the sidebar. If `guide.md` is absent from an incomplete checkout while `_toctree.yml` still contains `local: guide`, this publishes a generation without `guide.md`, preserves the dangling sidebar entry, and exits 0. Before publishing a full run, resolve every sidebar `local` to a planned Markdown page and fail closed if any are missing.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+    not and the previous generation still stands.
+    """
+    generation = publish.generation_id(tree)
+    # A generation is named after its contents, which is a claim about what should be on disk
+    # rather than proof that it is. Check before taking the shortcut, or a published generation
+    # that has been damaged would be recognised as "already correct" and never repaired.
+    if generation == current and not publish.verify_generation(root, generation, tree):
+        print(f"[translate] generation {generation} is already published and intact, nothing to do")
+    else:
+        bad = publish.write_generation(root, generation, tree)
```
mishig25: [P1] Repair a damaged live generation under a fresh directory. When `generation == current` but verification finds damage, this writes directly into the directory named by `CURRENT`. Files are then replaced one by one, so readers can observe a partial repair; unexpected extra files are not removed either, leaving the generation permanently unverifiable. Build and verify a new unreferenced repair generation, then promote its pointer only after the entire tree is intact.

src/doc_builder/translate/cache.py
```diff
@@ -0,0 +1,159 @@
...
+def atomic_write(path, text):
+    """Write a file by writing a temporary one and moving it into place.
+
+    Never a half-written file for someone else to read. Four places needed this -- the cache
+    index, a translated paragraph, the manifest and the published pointer -- and each had its
+    own copy of the same six lines.
+    """
+    path = Path(path)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    tmp = path.with_name(path.name + ".tmp")
```
mishig25: [P1] Give each atomic writer a unique temporary path. Every process writing the same destination also writes the same `path.tmp`. In a deterministic two-writer reproduction, writer A returned success, writer B raised `FileNotFoundError`, and `CURRENT` contained B's bytes. Use a unique temporary file in the destination directory, opened exclusively, before `os.replace`; pointer/manifest publication also needs per-package/language serialization or compare-and-swap semantics.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+    # removes anything not in the tree it was given, so pointing a three-page smoke run at the
+    # live folder would delete the other 737 pages and replace the sidebar with three entries.
+    # Nothing here touches the live tree or its manifest.
+    preview = bool(args.pages_file)
+    root = publish.lang_root(bucket, args.package, f"{args.lang}.preview" if preview else args.lang)
+    # What is published right now, for falling back on a page's last good translation and for
+    # checking the manifest against reality. None on a first run.
+    # The pointer is read once and carried, rather than re-read by each thing that needs it.
+    current = None if preview else publish.read_pointer(root)
+    read_dir = root if preview else publish.current_dir(root)
```
mishig25: [P1] Derive the read directory from the captured pointer. The comment says the pointer is read once, but `current_dir(root)` reads it again. If another publisher promotes between these calls, `current` can name generation A while `read_dir` points to B, so reconciliation and fallback use inconsistent snapshots. Build `read_dir` from the already captured `current`, then recheck or serialize before the no-op/promotion decision.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+        print(f"[translate] published generation {generation}, {len(tree)} file(s), {len(failed)} page(s) failed")
+
+    removed = publish.gc_generations(root, generation)
+    if removed:
+        print(f"[translate] removed {len(removed)} old generation(s): {removed}")
+
+    # The manifest goes last, and only records a generation that is actually published. If the
+    # run dies before this the manifest describes an older generation, disagrees with the
+    # pointer, and the next run rebuilds rather than trusting it.
+    publish.save_manifest(manifest_target, {**manifest, "generation": generation})
```
mishig25: [P1] Prevent stale publishers from overwriting the manifest. Promotion and manifest publication are not one serialized or conditional transaction: writer A can promote A, writer B can promote and record B, and then A can overwrite the manifest with A even though `CURRENT` names B. Serialize publishers, or make the manifest generation-specific and select it through the same atomic pointer.

src/doc_builder/translate/publish.py
```diff
@@ -0,0 +1,345 @@
...
+    """Delete old generations, keeping the newest few and never the published one.
+
+    Newest by modification time. `current` is the published generation, passed in rather than
+    re-read here -- the caller already knows it. Failures are only wasted space, so they are
+    reported and otherwise ignored.
+    """
+    base = Path(root) / GENERATIONS
+    if not base.is_dir():
+        return []
+    candidates = [p for p in base.iterdir() if p.is_dir() and p.name != current]
```
mishig25: [P1] Do not trust a caller's stale generation during GC. Another publisher can promote B after this caller captured A; this exclusion then protects A and treats B as a deletion candidate. I reproduced GC removing B while `CURRENT` still named it. Promotion and cleanup need synchronization, deletion must re-read and preserve the live pointer, and retention should be long enough for a build that already selected an older generation to finish syncing it.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+        print(f"[translate] ERROR no readable {TOCTREE} under {source_dir}; refusing to publish")
+        return 2
+
+    wanted = {}
+    for plan in plans.values():
+        wanted.update(plan.segments)
+    wanted.update(toc_keys)
+
+    known = cache.load_index()
+    missing = {k: v for k, v in wanted.items() if k not in known}
```
mishig25, line 375: [P1] Verify cache blobs before excluding segments from the request. Index membership does not guarantee the blob exists or can be read. `get_many` later drops a missing blob, the page can pass the coverage threshold with that paragraph in English, but the key remains indexed so subsequent warm runs never request it again and return success. Compute missing segments from readable cache values, or repair the index and retry/abort before publishing.

src/doc_builder/translate/pipeline.py
```diff
@@ -0,0 +1,578 @@
...
+        return None
+    # Sorted, not in order: Japanese word order differs from English, so a model that moves a
+    # marker to the other end of the sentence is doing its job -- only dropping, repeating or
+    # inventing one is wrong.
+    if sorted(placeholder_indices(text)) != sorted(placeholder_indices(source_text)):
+        return None
+    # Brackets it made up but did not number, like a lone `⟧`.
+    if validate.check_invented_brackets(source_text, text):
+        return None
+    return text
```
mishig25, line 253: [P1] Reject per-segment English echoes before counting coverage. An answer identical to `source_text` passes these checks and is counted as covered in `assemble_page`. Translating only a heading while echoing every body paragraph therefore reports 100% coverage, publishes an English body under the Japanese disclosure, and exits 0. Compare normalized non-placeholder prose here and treat an unchanged segment as untranslated before caching or coverage accounting.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+            return 1
+        tree[TOCTREE] = toctree_text
+
+    print(validate.summarize(results, args.warn_failure_rate))
+
+    # The same question asked of the model rather than of the pages. A run where most requests
+    # came back empty can still leave every page passing, because a page keeps its English and
+    # goes on looking well-formed.
+    attempted = len(missing)
+    if attempted >= SEGMENT_GATE_MIN_ATTEMPTS:
```
mishig25: [P1] Do not disable failure detection for small batches. Below 20 attempts, even 0/N model success skips this gate. A single new paragraph or sidebar title can fail while cached content keeps page/global failure rates at zero, so the English fallback is published with exit code 0. Small batches may remain non-blocking, but they still need a failure signal: record the fallback in the manifest and return 1, while retaining a separate threshold for aborting the entire publish.

src/doc_builder/translate/segment.py
```diff
@@ -0,0 +1,371 @@
...
+# saw the inner `badge.svg`. Deleting that `!` kept every marker in place and passed every check.
+#
+# Shared with validate.py, so what gets masked and what gets checked cannot drift apart. That
+# drift is what let a rewritten URL through last time.
+# Each alternative below is written so that only one of them can match at any position. With two
+# ways to read every `![x](y)` -- as one nested link, or as a `!`, a `[x]` and some loose
+# characters -- the engine tries every combination before giving up, and
+# `mask("![" + "![x](y)"*16)` never finished. So the plain-character branch refuses the `!` that
+# starts an image, and the bare-brackets branch refuses a `]` that a destination follows.
+LINK_LABEL = rf"(?:(?!!\[)[^\[\]\n]|!?\[[^\[\]\n]*\]{LINK_DEST}|\[[^\[\]\n]*\](?!\())*"
```
mishig25: [P1] Support soft line breaks inside Markdown link labels. This grammar forbids newlines, so a valid link such as `[technical\nreport](url)` is only half-masked. The model can remove its exposed opener and validation also discards link state at the newline, so the corruption passes. The current Transformers docs contain 24 such links across 19 pages. Allow soft line breaks while still stopping at actual block boundaries.

src/doc_builder/translate/validate.py
```diff
@@ -0,0 +1,332 @@
...
+            continue
+        if char == "!" and text.startswith("![", i):
+            stack.append(("image", i))
+            i += 2
+            continue
+        if char == "[":
+            stack.append(("link", i))
+            i += 1
+            continue
+        if char == "]" and i + 1 < n and text[i + 1] == "(":
```
mishig25: [P1] Invalidate nested openers on ordinary closing brackets. The stack is popped only for `](`. In `[huggingface_hub[cli]](url)`, deleting the nested `[` leaves an opener that is incorrectly paired with the later destination, so the target multiset stays unchanged even though CommonMark no longer produces the intended link. Track every closing bracket's nesting effect or use a Markdown-aware scanner.

src/doc_builder/translate/validate.py
```diff
@@ -0,0 +1,332 @@
...
+    without changing how many links there are. A count is the same number whether or not the
+    links still point anywhere near the right place, and turning an image into a link leaves
+    the count untouched too.
+    """
+    targets = list(scan_links(text))
+    for match in REF_LINK_RE.finditer(text):
+        targets.append(("ref", match.group("dest").lower()))
+    for match in REF_DEF_RE.finditer(text):
+        targets.append(("ref-def", match.group(1).lower()))
+    targets.extend(("url", m.group(0)) for m in BARE_URL_RE.finditer(text))
```
mishig25: [P1] Do not rediscover bare URLs from restored Japanese prose. Japanese commonly attaches particles without spaces. A correct restoration such as `https://github.com/deepseek-ai/DeepSeek-V3を参照してください` is tokenized here as one longer URL, producing a false lost/gained error even though the protected URL placeholder survived exactly. Compare known URL values through placeholder provenance instead of rerunning this permissive regex over translated prose.

src/doc_builder/translate/validate.py
```diff
@@ -0,0 +1,332 @@
...
+    """Every link in the text as (kind, destination), so two versions can be compared.
+
+    Destinations rather than a count, and the kind alongside, because both can be damaged
+    without changing how many links there are. A count is the same number whether or not the
+    links still point anywhere near the right place, and turning an image into a link leaves
+    the count untouched too.
+    """
+    targets = list(scan_links(text))
+    for match in REF_LINK_RE.finditer(text):
+        targets.append(("ref", match.group("dest").lower()))
```
mishig25: [P2] Preserve the image kind for reference-style links. `REF_LINK_RE` captures whether the source used `!`, but every match is recorded as `ref`. Consequently `![Diagram][fig]` can lose `!`, become `[Diagram][fig]`, and still pass with identical targets. Mask `![` together and record `ref-image` separately from `ref`.

src/doc_builder/translate/segment.py
```diff
@@ -0,0 +1,371 @@
...
+    (
+        "math_dollar",
+        re.compile(
+            r"(?<![\\$])\$(?:"
+            # no padding: `$x^2$`
+            r"(?![\s$])(?:[^$\n]|\n(?![ \t]*\n)){0,200}?(?<![\s\\])"
+            r"|"
+            # padded on both sides: `$ K_{\text{past}} $`
+            r"[ \t](?![\s$])(?:[^$\n]|\n(?![ \t]*\n)){0,200}?(?<![\s\\])[ \t]"
+            r")\$(?!\$)"
```
mishig25: [P2] Avoid interpreting price ranges as inline math. The closing delimiter may be followed by a digit, so `US$5 to US$10`, `$5-$10`, and `$5/$10` are matched as formulas such as `$5 to US$`. That silently shields ordinary price prose from translation, and round-trip checks cannot detect it. Tighten the closing-dollar context—at least disallow a following digit—and add price-range fixtures.

src/doc_builder/commands/translate.py
```diff
@@ -0,0 +1,607 @@
...
+        print(
+            f"[translate] ERROR {rate:.1%} of pages failed, above the {args.max_failure_rate:.0%} limit "
+            f"-- publishing nothing and leaving the existing translation alone"
+        )
+        return 2
+
+    if preview:
+        publish.write_tree(root, tree)
+        print(f"[translate] wrote {len(tree)} file(s) to the preview tree")
+        return 0
```
mishig25: [P2] Signal unsuccessful preview runs. Preview mode reaches this unconditional success return before the rejection-rate status, and the empty-page guard is disabled for previews. A pages file selecting nothing, or a preview with failed pages falling back to English, can therefore write an empty/degraded smoke tree and exit 0. Keep previews isolated, but return nonzero when no requested page exists or the warning threshold is exceeded.

.github/workflows/build_main_documentation.yml
```diff
@@ -186,6 +193,60 @@ jobs:
...
+            # putting them in place is a rename instead of copying 700 files. Cleared out first
+            # in case an earlier run was interrupted halfway through.
+            staged="$target.incoming"
+            rm -rf "$staged" "$staged.pointer"
+
+            # The translation job writes each run to a folder of its own under generations/ and
+            # then rewrites CURRENT to name the finished one. Read the pointer and take that
+            # folder, so we can only ever download a run that completed -- a job still writing,
+            # or one that died partway, is in a folder nothing points at.
+            uvx --from huggingface_hub hf sync "$BUCKET/$lang/CURRENT" "$staged.pointer"
```
mishig25, line 225: [P1] Seed the generation layout before enabling this consumer. The existing Transformers/Japanese bucket still uses the flat layout (98 Markdown files plus `_toctree.yml`) and has neither `CURRENT` nor `generations/`. Enabling `translated_languages` therefore makes this sync fail until a producer has successfully seeded the new format. Please coordinate and verify that migration step before rollout, or retain a temporary legacy fallback.

mishig25 (changes requested):
Thanks for the follow-up. The disclosure-aware fallback, direct nested-image/bare-URL/inline-math cases, empty and unreadable source guards, and normal versioned publish path are materially improved. The focused translation suite passes (162 passed, 5 skipped), and 3,772 current-corpus cases pass. However, I reproduced each inline issue below against this head; they can still publish an incomplete or English fallback tree, corrupt or delete the pointed generation under repair/overlap, or silently mishandle protected Markdown. I am therefore keeping this at changes requested. GitHub currently shows no check results for this head.

PR #816, opened 2026-08-18
Remove setup.py in favor of pyproject.toml

albertvillanova:
This PR removes `setup.py`, leaving `pyproject.toml` as the single source of packaging configuration.

### Motivation

Since #636, `pyproject.toml` declares a complete `[project]` table, which setuptools treats as authoritative for the package metadata. `setup.py` has been unused for metadata ever since, and it has drifted accordingly:

- `install_requires` still lists `black`, and `extras["quality"]` still pins `black~=22.0`, `isort` and `flake8`, all replaced by `ruff` in #636
- `extras["testing"]` still lists `requests`, replaced by `httpx` in #754

Keeping it around is actively misleading: #785 edited `setup.py` to fix a missing `ruff` dependency and had no effect at all, because the built metadata comes from `pyproject.toml` (see #815).

### Solution

While the `[project]` metadata was already coming from `pyproject.toml`, two settings in `setup.py` were still load-bearing, and both need to be migrated.

**1. `package_data`.** `setup.py` declared:

```python
package_data={"doc_builder": ["mock_deps/*.txt"]},
```

`MANIFEST.in` does not cover `mock_deps`, so without this declaration the mock-deps registry files are silently dropped from the wheel, and `doc-builder light-install <package> --check` then fails:

```
No mock-deps registry entry for `accelerate` (expected .../site-packages/doc_builder/mock_deps/accelerate.txt); do a full install instead.
```

It is therefore ported to `pyproject.toml` as `[tool.setuptools.package-data]`. `mock_deps` is the only directory under `src/doc_builder` holding non-Python files, so this is the only package data to declare.

**2. `is_doc_builder_repo`.** It identified the repo by reading the first line of `setup.py`, and it is load-bearing too: `locate_kit_folder` uses it as its second strategy to find the `kit` folder, after looking next to the installed module. In the documentation build workflows the package is installed non-editably and `doc-builder build` runs from the repo checkout, so the first strategy fails and this second one is what resolves `kit` locally. Removing `setup.py` without touching it would silently fall through to `get_cached_repo`, which clones `doc-builder` from GitHub, adding a network dependency to every documentation build.

It is switched to identify the repo from `pyproject.toml` instead. A plain string check is used rather than parsing TOML, to avoid depending on `tomllib`, which is only available from Python 3.11 while the project supports 3.10.

Verified by reproducing the workflow layout (non-editable install, `doc-builder` invoked from the repo checkout): `locate_kit_folder` still resolves to the local `kit` folder and performs no clone. With the previous `setup.py`-based check it returns `False` and would clone. No test covers this path.

With both migrated, building a wheel from a clean tree on this branch and from a clean tree on `main` produces an identical set of files, and the only metadata difference is that two legacy fields are no longer emitted:

```
- Home-page: https://github.com/huggingface/doc-builder
- Author: Hugging Face, Inc.
```

Both are already covered by their modern equivalents, which `pyproject.toml` declares and which are what PyPI displays:

```
Author-email: "Hugging Face, Inc." <docs@huggingface.co>
Project-URL: Homepage, https://github.com/huggingface/doc-builder
```

Finally, `setup.py` carried the release checklist in trailing comments. It is moved to `RELEASE.md` rather than dropped, with the version locations updated to `pyproject.toml` and the deprecated `python setup.py bdist_wheel`/`sdist` invocations replaced by `uv build`.

### Changes

- Remove `setup.py`
- Declare `mock_deps/*.txt` as package data in `pyproject.toml`
- Identify the `doc-builder` repo from `pyproject.toml` instead of `setup.py` in `is_doc_builder_repo`
- Move the release checklist from `setup.py` comments to `RELEASE.md`, updated for `pyproject.toml` and `uv build`

### Notes

The documentation build workflows check out `huggingface/doc-builder` without a `ref` and then run `git pull origin main` before `uv pip install .`, so they always install the current `main`, regardless of the commit a consumer pins the reusable workflow to. Pinning `build_main_documentation.yml@<sha>` freezes the workflow file only, not the `doc-builder` code it installs. This change therefore takes effect immediately for all consumers on their next build, and no consumer needs to bump anything.

This PR touches `pyproject.toml`, as does #815, but in a different section, so the two remain independent.


coyotte508 (commented):

This PR deletes `setup.py` (long unused for metadata since #636 and drifted out of sync), migrating its two remaining load-bearing pieces — the `mock_deps/*.txt` package-data declaration and the repo-detection used by `locate_kit_folder` — into `pyproject.toml`, and moving the release checklist into a new `RELEASE.md`. The PR description is exemplary and the two load-bearing migrations are correctly identified and handled.

**Findings**

- 🟡 **Lost "do not change" guard on the sentinel string** (`pyproject.toml` L2 / `src/doc_builder/utils.py` L124–127). The old `setup.py` opened with `# The line above is checked by some of the utilities in this repo, do not change it.` — that protection is gone. `is_doc_builder_repo` now depends on the exact byte sequence `name = "hf-doc-builder"`; a future TOML reformat (`name="hf-doc-builder"`, single quotes, etc.) would silently break the check, and the failure mode is nasty: `locate_kit_folder` falls through to `get_cached_repo()` and clones from GitHub on every doc build. Suggest adding a comment next to `name = "hf-doc-builder"` in `pyproject.toml`, and/or making the match whitespace/quote-tolerant (e.g. `re.search(r'name\s*=\s*["\']hf-doc-builder["\']', ...)`).
- 🟡 **No test coverage** for `is_doc_builder_repo`/`locate_kit_folder` (acknowledged in the PR). `tests/test_utils.py` exists and a unit test is cheap: assert `is_doc_builder_repo(repo_root)` is `True` for the actual repo root and `False` for an empty tmpdir / a tmpdir with an unrelated `pyproject.toml`. That would also catch the sentinel-drift scenario above.
- 🟡 **`setup.cfg` is left behind with the same kind of rot this PR targets**: dead `[isort]` (with `known_first_party = transformers`!) and `[flake8]` sections superseded by ruff in #636. The `[tool:pytest]` section is still live, so it can't just be deleted, but the stale sections could go here or in a follow-up — worth a note since the PR's stated goal is removing misleading packaging config.
- 🔵 `src/doc_builder/utils.py` L127: `read_text()` without `encoding="utf-8"` — uses the platform default (cp1252 on Windows). Harmless while `pyproject.toml` is ASCII, but `read_text(encoding="utf-8")` is safer. (The old code had the same flaw, so not a regression.)
- 🔵 `RELEASE.md` L29 faithfully carries over an outdated URL: `pip install -i https://testpypi.python.org/pypi hf-doc-builder` — TestPyPI moved to `https://test.pypi.org/simple/` years ago. Also, step 7 installs from real PyPI *before* the TestPyPI install, which reads oddly. Both pre-existing text, but a brand-new file is the natural moment to fix them.
- 🔵 `RELEASE.md` L20: `rm -rf dist build` — `uv build` doesn't produce a `build/` directory, so `build` in the cleanup is vestigial. Harmless.
- ❓ **Question:** any concern about *already-released* `hf-doc-builder` versions (whose installed `is_doc_builder_repo` still looks for `setup.py`) being run from inside an up-to-date repo checkout? They'd now fall through to the GitHub clone. Probably an acceptable, rare edge case — the CI workflows always install current `main` — but worth confirming it's a non-goal.

Verified against the workspace: `pyproject.toml` already supersedes all of `setup.py`'s stale deps (`black`, `isort`, `flake8`, `requests`); `MANIFEST.in` indeed does not cover `mock_deps`, so the `[tool.setuptools.package-data]` addition is genuinely needed; the release checklist's file references (`src/doc_builder/__init__.py` L20 and `pyproject.toml` L3 both hold `0.6.0.dev0`) are accurate; and a repo-wide grep confirms no other file references `setup.py`.

**Verdict:** Solid, well-motivated cleanup with both load-bearing behaviors correctly migrated — approve once the sentinel-string fragility is guarded (comment or tolerant match), ideally with a small `is_doc_builder_repo` test.

albertvillanova:
Thanksfor the review @coyotte508, this is a good catch on the sentinel string. Addressed in 6bc1117.

**Sentinel fragility.** Agreed, and the failure mode you describe is exactly why it matters: a silent fall-through to `get_cached_repo()` means a network clone on every doc build. `is_doc_builder_repo` now uses an anchored, quote and whitespace tolerant match, and `pyproject.toml` carries a guard comment in place of the one that died with `setup.py`:

```python
return re.search(r"""^name\s*=\s*["']hf-doc-builder["']""", content, flags=re.MULTILINE) is not None
```

Verified against the drift cases: `name='hf-doc-builder'` and `name=  'hf-doc-builder'` now match, while the old substring check returned `False` for both. A commented-out line and the string embedded in another value still correctly return `False`, so the anchor does not over-match.

**Test coverage.** Added `test_is_doc_builder_repo` to `tests/test_utils.py`, covering the repo root, an empty tmpdir, an unrelated `pyproject.toml`, and the quoting and spacing variants, so sentinel drift is now caught by the suite.

**Encoding.** Fixed, `read_text(encoding="utf-8")`.

**`rm -rf dist build`.** Confirmed and removed. Running `uv build` on a clean `git archive` leaves only `dist/`; the `removing build/bdist...` line in the output is setuptools cleaning up inside its isolated build directory.

**TestPyPI URL.** Fixed to `https://test.pypi.org/simple/`. I left the step ordering as it was: installing from real PyPI before TestPyPI does read oddly, but reordering the release procedure felt like more than a faithful relocation of the checklist should decide unilaterally. Happy to change it if you prefer.

**`setup.cfg`.** Agreed the `[isort]` and `[flake8]` sections are dead, superseded by ruff in #636, and `known_first_party = transformers` is a nice fossil. I have left it out of this PR since it is lint configuration rather than packaging metadata, and `[tool:pytest]` is still live so the file cannot simply go. Happy to do it here instead of a follow-up, just say which you prefer.

**On already-released versions.** Confirming it is a non-goal. An installed `hf-doc-builder` from before this change still looks for `setup.py`, so running it from inside an up-to-date checkout falls through to the cached clone. It still works, just with a clone on first use, and the doc build workflows always install current `main`, so the path is not exercised there.

coyotte508 (approved):

mishig25 (approved):

PR #817, opened 2026-08-19
fix(kit): restore whitespace eaten by \(...\) and $$ KaTeX markers

### LINKED ISSUE #809: Inline `\(...\)` math swallows the preceding whitespace; the custom delimiter regexes also diverge from standard KaTeX
The Markdown preprocessor renders math by extracting it with hand-written regexes (`markKatex` in `kit/preprocessors/mdsvex/index.js`) before Markdown runs, then rendering the stashed TeX with `katex.renderToString`. There is a concrete bug in that extraction, plus a broader issue: the custom delimiter rules diverge from standard KaTeX in ways that silently bite contributors.

## 1. Bug: inline `\(...\)` eats the whitespace before it

```js
const REGEX_LATEX_INLINE = /\s\\\\\(([\s\S]+?)\\\\\)/g;
// ...
.replace(REGEX_LATEX_INLINE, (_, tex) => {
  const marker = `KATEXPARSE${counter++}MARKER`;
  markedKatex[marker] = { tex, displayMode: false };
  return marker;            // the leading \s that was matched is never restored
});
```

The regex matches a leading `\s`, but the replacement returns only `marker`, so that whitespace character is consumed and dropped. Two visible effects:

- **A space between text and an inline formula disappears:** `... one step \(S_{t+1}\)` renders as `one stepS_{t+1}` (glued).
- **Standalone equations run together:** when the eaten character is the newline separating two `\(...\)` lines, the blank-line paragraph break collapses and consecutive equations jam onto one line.

The `$...$` branch a few lines below handles this correctly — it captures the surrounding whitespace and re-emits it (`return spaceBefore + marker + spaceAfter`). The `\(` branch (and the `$$` branch, which eats a leading `\n`) do not.

**Fix** — assert the boundary without consuming it, or capture-and-restore:

```js
const REGEX_LATEX_INLINE = /(?<=\s)\\\\\(([\s\S]+?)\\\\\)/g;   // lookbehind
// or: /(\s)\\\\\(([\s\S]+?)\\\\\)/g  ->  return spaceBefore + marker;
```

## 2. The delimiter rules are inconsistent

| Syntax | Rule in `markKatex` |
|---|---|
| `\(...\)` inline | requires (and eats) one leading whitespace |
| `$...$` inline | requires whitespace on **both** sides |
| `$$...$$` display | requires (and eats) a leading newline |
| `\[...\]` display | **not handled at all** |

There is no consistent mental model, and the natural display counterpart of `\(...\)` — `\[...\]` — is not supported (you must use `$$`).

## 3. The real cost: standard KaTeX syntax silently fails

Contributors will reach for standard KaTeX syntax, but much of it does not render here — it silently becomes literal text instead of math. Examples that fail:

- `$x$` right before punctuation or markup, e.g. `$x$.` or `**$x$**`
- `\[ ... \]` display math — not handled at all
- `\(...\)` at the start of a line — no leading space for the regex to match

We hit this while fixing one page in the Deep RL course: using standard `$...$` and `\[...\]` silently broke ~20 formulas, and it took reading this file to find out why.

**Suggestion:** handle math with the standard toolchain — `remark-math` + `rehype-katex`, or KaTeX's `auto-render` — instead of hand-written regexes, so standard syntax just works and the issues above go away together.

## Where it shows

Live example on the Deep RL course: https://huggingface.co/learn/deep-rl-course/unit2/mc-vs-td (source `units/en/unit2/mc-vs-td.mdx`). Before/after below:

Image

hungnnvidia:
## Summary
- Fixes the KaTeX preprocessor dropping the leading whitespace that `\(...\)` and `$$...$$` require to match, which glued formulas to the preceding text and collapsed consecutive `\( ... \)` lines (reported in #809).
- `$...$` already captured and restored surrounding spaces; the other two delimiters now do the same.
- Adds `node --test` coverage for `markKatex` (`npm test` in `kit/`).

This does **not** take on the rest of #809 (switching to `remark-math` / supporting `\[...\]` / relaxing `$...$` punctuation boundaries). Those are larger syntax-policy changes.

## Test plan
- [x] `cd kit && npm test`
- [x] `uv run --python 3.10 python -m pytest -n 1 --dist=loadfile -q ./tests/` (138 passed; unrelated to this change)
- [ ] Confirm Deep RL course page (`units/en/unit2/mc-vs-td`) no longer renders `one stepS_{t+1}`

Fixes #809 (whitespace-eating part)

### DIFF
--- a/kit/package.json
+++ b/kit/package.json
@@ -8,6 +8,7 @@
 		"preview": "vite preview",
 		"check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
 		"check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
+		"test": "node --test preprocessors/mdsvex/markKatex.test.js",
 		"lint": "prettier --check --plugin prettier-plugin-svelte .",
 		"format": "prettier --write --plugin prettier-plugin-svelte .",
 		"update-inference-providers": "npm ci && npm i @huggingface/inference@latest && npm i @huggingface/tasks@latest"
--- a/kit/preprocessors/mdsvex/index.js
+++ b/kit/preprocessors/mdsvex/index.js
@@ -97,24 +97,28 @@ export const mdsvexPreprocess = {
  * @param {string} content
  * @param {Record<any, any>} markedKatex
  */
-function markKatex(content, markedKatex) {
-	const REGEX_LATEX_DISPLAY = /\n\$\$([\s\S]+?)\$\$/g;
-	const REGEX_LATEX_INLINE = /\s\\\\\(([\s\S]+?)\\\\\)/g;
+export function markKatex(content, markedKatex) {
+	// Capture-and-restore the required boundary whitespace so formulas are not
+	// glued to the preceding text (and consecutive `\( ... \)` lines don't
+	// collapse into one paragraph). `$...$` already does this; `\(...\)` and
+	// `$$...$$` previously consumed the leading `\s`/`\n` and dropped it.
+	const REGEX_LATEX_DISPLAY = /(\n)\$\$([\s\S]+?)\$\$/g;
+	const REGEX_LATEX_INLINE = /(\s)\\\\\(([\s\S]+?)\\\\\)/g;
 	// Match $...$ with whitespace boundaries to avoid matching in HTML/code
 	const REGEX_LATEX_INLINE_DOLLAR = /(\s)(\$)([^$\n`<>]+?)(\$)(\s)/g;
 	let counter = 0;
 	return content
-		.replace(REGEX_LATEX_DISPLAY, (_, tex) => {
+		.replace(REGEX_LATEX_DISPLAY, (_, newline, tex) => {
 			const displayMode = true;
 			const marker = `KATEXPARSE${counter++}MARKER`;
 			markedKatex[marker] = { tex, displayMode };
-			return marker;
+			return newline + marker;
 		})
-		.replace(REGEX_LATEX_INLINE, (_, tex) => {
+		.replace(REGEX_LATEX_INLINE, (_, space, tex) => {
 			const displayMode = false;
 			const marker = `KATEXPARSE${counter++}MARKER`;
 			markedKatex[marker] = { tex, displayMode };
-			return marker;
+			return space + marker;
 		})
 		.replace(
 			REGEX_LATEX_INLINE_DOLLAR,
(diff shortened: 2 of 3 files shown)


coyotte508 (commented):

This PR fixes a real whitespace-eating bug in `markKatex` (`kit/preprocessors/mdsvex/index.js`) by capturing the mandatory leading `\n`/`\s` before `$$...$$` and `\\(...\\)` and re-emitting it in the replacement — exactly the pattern the `$...$` branch already used — and adds the first `node --test` unit tests for the preprocessor. The change is minimal, correct, and well-scoped (it deliberately defers the larger `remark-math`/`\[...\]` questions from #809).

**Findings**

- 🟡 **New test suite isn't wired into CI.** `.github/workflows/lint-svelte-kit.yml` runs `npm run lint` and `npm run check` (lines 29–35) but never `npm test`, so the new `kit/package.json` `"test"` script only runs locally. Suggest adding an `npm test` step (with `working-directory: kit`) so the regression test actually guards against regressions.
- 🟡 **Test imports a module with heavy side effects.** `markKatex.test.js` imports `./index.js`, whose module scope calls `hljs.addPlugin(...)`, constructs `mdsvex(...)`, and does a top-level `await findSvelteComponentNames("./src/lib")` (`index.js:166–167`) — a filesystem scan relative to `process.cwd()`. The test therefore requires `npm ci` in `kit/` and only works when run from the `kit/` directory (`npm test` guarantees that, but bare `node --test` from repo root would fail). Consider extracting `markKatex`/`renderKatex` into a small pure module (e.g. `katex.js`) so the unit test has no import side effects.
- 🔵 **Known limitations left as-is (fine, but worth a line in the PR body):** math at the very start of the content still won't match (both regexes require a preceding `\n`/`\s`, `index.js:105–106`), and `$...$` still consumes its *trailing* space from the scanner's perspective, so `$a$ $b$` with a single separating space only marks the first. Both pre-existing, not regressions.
- 🔵 The display regex comment says "leading `\s`/`\n`" collectively; the display marker specifically requires `\n` only. Trivial wording nit in the new comment block (`index.js:98–101`).

Correctness itself checks out: tracing the replacement flow in `index.js:100–128`, restoring the captured char cannot create spurious matches in the two subsequent `.replace` passes (markers contain no `$`, `\`, or whitespace), and the four tests cover the reported repro (`one stepS_{t+1}`), consecutive-line collapse, `$$` paragraph boundaries, and the untouched `$...$` path. The two-backslash test inputs correctly mirror what `convert_rst_to_mdx`/notebook conversion emits. Style (tabs, double quotes) matches the file.

**Verdict:** LGTM with minor suggestions — the fix is correct and consistent with the existing `$...$` handling; the only meaningful gap is that CI doesn't run the new tests.

PR #818, opened 2026-08-19
Ship a pre-commit hook for doc-builder style

albertvillanova:
This PR adds a `.pre-commit-hooks.yaml` manifest, so that `doc-builder style` can be used as a regular [pre-commit](https://pre-commit.com) hook by the repos that consume `doc-builder`.

### Motivation

Repos that want to enforce doc styling currently have to declare a `repo: local` hook and install `doc-builder` from a git URL pinned inside `additional_dependencies`. For example, in `trl`:

```yaml
- repo: local
  hooks:
    - id: doc-builder-style
      language: python
      entry: doc-builder style trl tests docs/source --max_len 119
      additional_dependencies: ["git+https://github.com/huggingface/doc-builder@<sha>"]
      pass_filenames: false
      types_or: [python, markdown, rst]
```

This works, but the pinned revision is invisible to the tooling that would otherwise keep it current: `pre-commit autoupdate` only rewrites the `rev` of remote hook repos, and Dependabot's `pre-commit` ecosystem tracks the same field. Neither looks inside `additional_dependencies`, so the pin can only be moved by hand and silently goes stale.

Shipping a manifest here lets consumers use the standard form instead, where `rev` is a first-class field that both tools can bump:

```yaml
- repo: https://github.com/huggingface/doc-builder
  rev: <rev>
  hooks:
    - id: doc-builder-style
```

It also removes the need for consumers to know about `doc-builder`'s own dependencies, now that `ruff` is declared as one (#815).

### Solution

The hook runs `doc-builder style` as a `language: python` hook, and relies on pre-commit's defaults for the rest.

`types_or` is `[python, mdx]`, which is exactly what `style_doc_files` handles: any other extension only triggers an `Ignoring <file> because it's not a py or an mdx file or a folder` warning. Note this is narrower than the `[python, markdown, rst]` used by the local hooks currently in the wild, which is harmless there only because they also set `pass_filenames: false`.

`pass_filenames` is left at its default, so pre-commit passes the files that are about to be committed and the hook is incremental, rather than restyling the whole tree on every run. `--max_len` is not passed either, since `119` is already the default; consumers can override both behaviours with `args` and `files`.

Verified with `ruff` absent from `PATH`, to make sure the hook is self-sufficient:

- `pre-commit validate-manifest .pre-commit-hooks.yaml` passes
- `pre-commit try-repo` restyles a Python docstring and reformats its code example, leaves `.md` and `.rst` files untouched, and exits non-zero because files were modified, like any other formatter
- a second run passes, so the hook is idempotent
- the documented `repo`/`rev` form, with `args` and a `files` filter, restyles the included folder and skips the excluded one

### Changes

- Add `.pre-commit-hooks.yaml` declaring the `doc-builder-style` hook
- Document `doc-builder style` and the hook in a new `Doc styling` section of the README

### Note

The README example asks for a `rev` rather than showing a concrete one, since the latest release, `v0.5.0`, predates this manifest and would fail with a missing-hook error. It can be replaced by a version once a release includes the hook.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: doc-builder-style
+  name: Check style with doc-builder
+  description: Style docstrings in Python files and code examples in mdx files.
+  entry: doc-builder style
+  language: python
+  types_or: [python, mdx]
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ This is the package we use to build the documentation of our Hugging Face repos.
   * [Installation](#installation)
   * [Previewing](#previewing)
   * [Doc building](#doc-building)
+  * [Doc styling](#doc-styling)
+    + [Using the pre-commit hook](#using-the-pre-commit-hook)
   * [Writing in notebooks](#writing-in-notebooks)
   * [Templates for GitHub Actions](#templates-for-github-actions)
     + [Enabling multilingual documentation](#enabling-multilingual-documentation)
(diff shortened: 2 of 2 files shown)


coyotte508 (commented):

This PR adds a 6-line `.pre-commit-hooks.yaml` manifest exposing `doc-builder style` as a standard pre-commit hook (so consumers can use `repo:`/`rev:` instead of a `repo: local` hook), plus a new "Doc styling" README section documenting the command and the hook. The design choices are sound and well-argued: `types_or: [python, mdx]` matches exactly what `style_doc_files` handles (`src/doc_builder/style_doc.py:533-566`), filenames pass through to the `files` positional of the CLI (`src/doc_builder/commands/style.py:38`), `--max_len 119` is already the default (`style.py:40`), and relying on pre-commit's modified-files detection for failure is how formatters like `ruff-format` behave.

**Findings**

- 🟡 **Consider `minimum_pre_commit_version` in the manifest** (`.pre-commit-hooks.yaml`). `types_or` requires pre-commit ≥ 2.9.0, and the `mdx` identify tag only exists in `identify` ≥ 2.5.19 (early 2023). On older setups the hook fails with an opaque manifest-validation error rather than a clear version message. Adding `minimum_pre_commit_version: "2.9.0"` (or higher) is one line and makes the failure mode self-explanatory. Not blocking — anything from the last ~2 years works fine.
- 🟡 **No automated guard on the manifest.** Nothing in CI validates `.pre-commit-hooks.yaml`, so a future typo (e.g., renaming the console script) would only surface in consumer repos. A `pre-commit validate-manifest .pre-commit-hooks.yaml` step in the existing `quality` workflow, or a tiny test asserting the hook id/entry, would be cheap insurance. The manual verification described in the PR is thorough but one-time.
- 🔵 identify's `python` tag also matches extensionless scripts via shebang (`#!/usr/bin/env python`). Those get passed to the hook and hit the `Ignoring {file} because it's not a py or an mdx file` warning at `style_doc.py:566` — harmless (exit 0, no modification), just warning noise.
- 🔵 The second README YAML snippet (the `args`/`files` example) starts at hook-list indentation and isn't standalone-valid YAML. It reads fine as a continuation of the block above, but a lead-in like "extending the config above:" would remove any ambiguity.
- 🔵 The hook `description` ("Style docstrings in Python files and code examples in mdx files") slightly undersells it — `style_file_docstrings` also reformats code examples *inside* Python docstrings, which is arguably the main feature. Minor wording only.
- ❓ **Question:** several HF repos now use `.md` (not `.mdx`) for doc sources, and `doc-builder style` ignores `.md` entirely (`style_doc.py:540/553/566`), so the hook can't cover them even with a `files` override. Is `.md` support for `style_doc_files` (and then `markdown` in `types_or`) on the roadmap, or intentionally out of scope? Also: when a code block can't be parsed by ruff, `style_doc_files` raises `ValueError` (`style_doc.py:569`), so the hook fails with a Python traceback rather than a clean message — pre-existing CLI behavior, but worth knowing that's the UX consumers will see.

**Verdict:** LGTM — small, correct, well-documented; the suggestions above (minimum version pin, manifest validation in CI) are nice-to-haves, not blockers.

PR #819, opened 2026-08-20
Escape HTML and strip ANSI codes in notebook-to-mdx text outputs

sergiopaniego:
# What does this PR do?

Fixes the doc build for repos using `convert_notebooks: true` (the cookbook), broken for every PR since the kit upgrade to Svelte 5 (#792).

`notebook_to_mdx` wrapped cell text outputs in `<pre>` without escaping, so tag-like text in an output reached the Svelte compiler as raw markup. Example from `benchmarking_tgi.ipynb` (the committed output of `text-generation-launcher -h`):

```
[vite-plugin-svelte:compile] src/routes/benchmarking_tgi/+page.svelte (89:26):
Expected a valid element or component name.
 89 |        --model-id <MODEL_ID>
```

Svelte 4 tolerated these as unknown elements, Svelte 5 rejects them, and 41 of the 71 English cookbook notebooks have outputs with tag-like text (`<MODEL_ID>`, `<https://...>`, object reprs like `<transformers.Trainer object at 0x...>`), so this can't reasonably be fixed notebook-side. Curly braces were already handled by the kit's `onHtml` preprocessor, which is why JSON in outputs never broke.

Changes:
- HTML-escape the output text before wrapping it in `<pre>` (`html.escape(..., quote=False)`).
- Strip ANSI escape sequences (colors, cursor control like `\x1b[?25h`), which rendered as garbage characters inside the `<pre>` blocks.
- Add tests for both behaviours.

Verified by converting all 71 cookbook `en` notebooks with the patched command: no tag-like text and no ANSI sequences remain inside any generated `<pre>` block, and `benchmarking_tgi.md` now renders `--model-id &lt;MODEL_ID&gt;` correctly.

Once merged, re-running the failed cookbook checks should go green with no cookbook-side change (the reusable PR-doc workflow installs doc-builder from `main`).

### DIFF
--- a/src/doc_builder/commands/notebook_to_mdx.py
+++ b/src/doc_builder/commands/notebook_to_mdx.py
@@ -14,7 +14,9 @@
 
 import argparse
 import base64
+import html
 import io
+import re
 from pathlib import Path
 
 import nbformat
@@ -22,6 +24,23 @@ from PIL import Image
 from tqdm import tqdm
 
 from ..style_doc import format_code_example
+
+
+# CSI sequences (colors, cursor control like `\x1b[?25h`) and OSC sequences (e.g. terminal titles).
+_ANSI_ESCAPE_RE = re.compile(r"\x1b(?:\[[0-?]*[ -/]*[@-~]|\][^\x07\x1b]*(?:\x07|\x1b\\)?)")
+
+
+def text_output_to_pre(output):
+    """
+    Wrap a cell's text output in a `<pre>` block, made safe for the MDX/Svelte pipeline.
+
+    ANSI color codes are stripped (they would render as garbage characters), and the text is
+    HTML-escaped: anything tag-like in an output (e.g. `<MODEL_ID>` in a CLI --help dump, or an
+    object repr like `<transformers.Trainer object at 0x...>`) would otherwise reach the Svelte
+    compiler as raw markup, which Svelte 5 rejects as an invalid element or component name.
+    """
+    output = _ANSI_ESCAPE_RE.sub("", output)
+    return f"<pre>\n{html.escape(output.strip(), quote=False)}\n</pre>"
 
 
 def png_to_jpeg(png_base64):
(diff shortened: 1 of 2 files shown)


coyotte508 (commented):

This PR fixes cookbook doc builds broken by Svelte 5's strict tag parsing, by HTML-escaping notebook cell text outputs and stripping ANSI escape sequences before wrapping them in `<pre>` in `notebook_to_mdx`. The change is small, well-scoped, correctly interacts with the kit's downstream preprocessors, and comes with a new test file.

**Findings**

- 🔵 **Downstream interaction verified — no double-escaping.** In the kit pipeline, `renderSvelteChars` (which decodes `&amp;lt;` → `<`) only runs on katex, highlighted code blocks, and docstrings — not raw `<pre>` HTML nodes — and `renderCode`'s `&amp;` → `&` pass only touches `<code>` elements. `onHtml` in `kit/preprocessors/mdsvex/index.js` (line 314) round-trips the `<pre>` block through cheerio, which preserves `&lt;` entities and escapes `{` as the PR body claims. So `html.escape(..., quote=False)` at `src/doc_builder/commands/notebook_to_mdx.py:41` is the right escaping level here, even though `convert_rst_to_mdx.py:162` uses the different `&amp;lt;` convention (that convention is for MDX text, not HTML nodes — no inconsistency in practice).
- 🟡 **Carriage returns from progress bars are not handled.** tqdm/pip-style outputs use `\r` to redraw a line; after ANSI stripping, `<pre>\nloading: 10%\rloading: 100%\n</pre>` still contains the CRs, which HTML renders as line breaks — so every intermediate progress frame appears as its own line. This is pre-existing behavior, but since the PR already normalizes terminal noise it's a natural companion fix (e.g. keep only the text after the last `\r` per line). Fine as a follow-up.
- 🔵 **ANSI regex misses non-CSI/OSC escapes.** `_ANSI_ESCAPE_RE` (notebook_to_mdx.py:29) handles CSI and OSC (BEL- and ST-terminated, including hyperlink sequences), but not two-character escapes like `\x1b(B` (charset selection) or `\x1bM`, nor backspace `\x08`. These are rare in notebook outputs; acceptable as-is.
- 🔵 **Duplicate ANSI-stripping logic in the codebase.** `src/doc_builder/style_doc.py:155` already strips ANSI color codes with a simpler regex (`\x1b\[[0-9;]*m`). Not worth blocking on, but the new, more complete `_ANSI_ESCAPE_RE` could live in a shared util and be reused there.
- 🔵 **Test coverage is good but the `text/plain` branch is untested.** `tests/test_notebook_to_mdx.py` covers escaping, ANSI stripping, and an end-to-end `stream` output (with the required `name: "stdout"` field — correctly matching the filter at notebook_to_mdx.py:45). The `elif "text/plain"` branch goes through the same helper, so risk is low; a one-liner test for an `execute_result`-style output would make it airtight. (Helper logic and both unit-test expectations re-executed locally — they pass.)
- 🔵 **Faithfulness of already-escaped output.** An output that literally prints `&lt;` becomes `&amp;lt;` and renders as the literal text `&lt;` — which is the *correct* faithful rendering of what the cell printed, not a bug. Noting it only so you can confirm this is intended.

**Question:** was skipping `\r` handling deliberate (out of scope), or just not encountered in the cookbook notebooks you verified against?

**Verdict:** LGTM — correct, well-tested, and verified against the kit's preprocessing pipeline; only optional polish items above. ✅

PR #820, opened 2026-08-20
Check anchors in links

stevhliu:
The link checker removes the URL fragments before checking the path, so it only knows the file exists and it doesn't know the anchor exists (for example, `[install](#how-to-install)` would pass even if `(#how-to-install)` was removed or renamed) This includes the anchors in the check.

Also includes some cases where a link is inside a code block and other valid patterns.

cc @sayakpaul for vis :)

### DIFF
--- a/src/doc_builder/check_links.py
+++ b/src/doc_builder/check_links.py
@@ -1677,12 +1677,6 @@ def _resolve_autodoc_anchors(all_files: list[Path], package_name: str | None) ->
     if not package_name:
         return {}
 
-    try:
-        package = importlib.import_module(package_name)
-        from .autodoc import find_object_in_package, get_shortest_path
-    except Exception:
-        return {}
-
     object_names = set()
     for file_path in all_files:
         try:
@@ -1692,13 +1686,20 @@ def _resolve_autodoc_anchors(all_files: list[Path], package_name: str | None) ->
         object_names.update(
             match.group(1) for line in content.splitlines() if (match := _re_autodoc.match(line)) is not None
         )
+    if not object_names:
+        return {}
+
+    try:
+        package = importlib.import_module(package_name)
+        from .autodoc import find_object_in_package, get_shortest_path
+    except Exception:
+        return {}
 
     resolved_anchors = {}
     for object_name in object_names:
         try:
             obj = find_object_in_package(object_name, package)
             if obj is None:
-                resolved_anchors[object_name] = None
                 continue
             shortest_path = get_shortest_path(obj, package)
             resolved_anchors[object_name] = "None" if shortest_path is None else shortest_path
--- a/src/doc_builder/check_links.py
+++ b/src/doc_builder/check_links.py
@@ -37,7 +37,7 @@ except ImportError:
 _re_md_link = re.compile(r"!?\[([^\]]*)\]\(([^)]+)\)")
 _re_heading = re.compile(r"^\s{0,3}#{1,6}\s+(.+?)\s*$")
 _re_custom_anchor = re.compile(r"(?:\[\[([^\]]+)\]\]|\[\s+([^\]]*?)\s+\])\s*$")
-_re_html_anchor = re.compile(r'''\b(?:id|name)\s*=\s*(?:"([^"]+)"|'([^']+)')''', re.IGNORECASE)
+_re_html_anchor = re.compile(r"""\b(?:id|name)\s*=\s*(?:"([^"]+)"|'([^']+)')""", re.IGNORECASE)
 _re_fence = re.compile(r"^\s{0,3}(`{3,}|~{3,})")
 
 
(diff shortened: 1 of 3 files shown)


coyotte508 (commented):

Extends the `doc-builder check-links` checker to validate `#fragment` anchors against target-file headings/IDs, skip links inside fenced/inline code, handle angle-bracketed URLs and root-relative paths, and fixes extension guessing for dotted page names (e.g. `t5v1.1`). Comes with a new, reasonable pytest test file.

**Findings**

- 🔴 **`[[autodoc]]`-generated anchors are not recognized** (`src/doc_builder/check_links.py`, `extract_anchors`). API pages get most of their anchors at build time: `autodoc` emits `transformers.Trainer`, `transformers.Trainer.train`, per-parameter anchors, etc., and `build_doc.py:159` even *appends* `[[<object_anchor>]]` to the preceding heading, replacing its slug on the published site. Since `extract_anchors` only reads the source markdown, links like `(trainer#transformers.Trainer)` or `(#transformers.Trainer.train)` will be reported as broken (false positives), while links to the source heading's slug (e.g. `#trainer`) pass even though they 404 on the rendered site (false negatives). For repos like `transformers`/`diffusers` this will make CI noisy. Suggest parsing `[[autodoc]] obj` lines (plus their `- method` list items) into the anchor set, or at minimum skipping anchor validation for files containing autodoc directives. Question: was this run against a large real docs tree (e.g. transformers) to gauge the false-positive rate?
- 🟡 **URL-encoded fragments won't match.** Translated docs (ko/zh/ja) have non-ASCII heading anchors that are often linked percent-encoded (`#%ED%9B%88%EB%A0%A8...`). `split_link_url` should `urllib.parse.unquote` the fragment before comparison. (The Unicode slug itself is handled correctly — `_heading_anchor("훈련 방법") == "훈련-방법"`, matching the frontend's `[^\p{L}\p{N}-]` rule in `kit/preprocessors/mdsvex/index.js:213-217`.)
- 🟡 **Anchor cache is per source file** (`anchor_cache` local to `check_file_links`). Hot targets (e.g. an index page linked from hundreds of files) get re-read and re-parsed once per referencing file per worker. A module-level `functools.lru_cache` on `extract_anchors` would keep the "fast link checker" promise on large repos.
- 🟡 **Fence-tracking state machine is duplicated** in `extract_anchors` and `check_file_links` — consider extracting a shared helper. Also, neither enforces that a closing fence be at least as long as the opener (CommonMark), and fences indented ≥4 spaces (common inside list items) aren't detected by `_re_fence`, so links/anchors inside them are still processed.
- 🟡 **Slug is computed from raw markdown, frontend uses rendered text.** Mostly aligned (backticks/bold/Unicode all check out), but a heading containing a link diverges: `## See [the docs](./page)` → checker computes `see-the-docspagemd`, site produces `see-the-docs`. Rare, but worth a comment or fix.
- 🟡 **Unterminated inline-code span suppresses the rest of the line.** `_inline_code_ranges` extends an unmatched backtick run to EOL, so one stray backtick before a link silently exempts it from checking (false negative). Consider treating unmatched runs as literal text instead.
- 🔵 `resolve_link_path` calls `is_anchor_only(original_link_url)` on the *un-stripped* URL, so an angle-bracketed anchor-only link `(<#foo>)` returns `None` and skips validation — inconsistent with the stripping done in `split_link_url`/`is_external_link`.
- 🔵 Setext headings (`Title` / `=====`) aren't matched by `_re_heading`; low risk given HF docs use ATX, but would be a false positive if encountered.
- 🔵 `tests/test_check_links.py` is missing the Apache license header that every other test file in `tests/` carries.
- 🔵 Root-relative `/docs/...` links are now counted in `total_links` but never validated — that's fine (and it actually fixes a pre-existing bug where `/`-rooted links resolved against the filesystem root and were flagged broken), but a docstring note on the counting semantics would help.

Good parts worth calling out plainly: the `find_target_file` rewrite correctly fixes dotted extensionless names (`t5v1.1`), the custom-anchor regex handles both `[[x]]` and `[ x ]` exactly as the frontend does, the code-block skipping removes a real class of false positives, and the tests are pytest-idiomatic with sensible coverage of the new paths (missing: autodoc pages, `.html` links, query strings, encoded fragments).

**Verdict:** Well-built and in the right direction, but the missing `[[autodoc]]` anchor handling will likely cause false CI failures on the main consumer repos — worth addressing (or adding an opt-out) before merge.

mishig25 (approved):

mishig25:
Follow-up commit 894d5d3 tightens autodoc resolution: unresolved objects now retain the conservative source-based fallback instead of being treated as definitively missing, and the documented package is not imported when the docs contain no autodoc directives. Regression coverage was added. Validation passes with 47 focused tests, 185 full-suite tests, and Ruff checks; the Transformers corpus remains stable at 86 existing findings across 737 files.

